Issue #11, opened 2024-03-27
AI generated documentation for Nox


nestordemeure:
Hi!

Would you be interested in a PR adding AI-generated documentation to the types, traits, and functions in Nox? It would be manually checked by a human (me) who has knowledge of Jax and Rust but not this particular codebase.

I have run some quick tests and the output seem serviceable, definitely not perfect but a good starting point to be tweaked by people familiar with the codebase before merging.

akhilles:
Hey @nestordemeure, we're certainly open to it but it depends heavily on the quality of the generated comments. Could we start with a small submodule and then review it?

My general preference for code comments is to not be overly verbose wrt. information that can easily be inferred from the function signature. Comments that explain complex/unintuitive behavior are the most welcome.

nestordemeure:
Do you have a submodule you would recommend? I could then make a PR to have comments to discuss.

sphw:
> Do you have a submodule you would recommend? I could then make a PR to have comments to discuss.

Could you try on the `tensor` module?

nestordemeure:
Will do and keep you posted once the PR is up :)

nestordemeure:
I think the `tensor.rs` trial was a success. My suggestion would be to keep going one file at a time (that way you have time to review the comments without being overwhelmed). In which case, in which order would you like me to cover the Nox files?

akhilles:
1 PR for the rest of the module is fine, and somewhat preferable. Reviewing it isn't a concern.

nestordemeure:
I have added a PR covering the full Nox library. That was a *lot* of comments and will definitely take some time to review but it looks like an acceptable basis to document the code.

akhilles:
> I have added a PR covering the full Nox library. That was a _lot_ of comments and will definitely take some time to review but it looks like an acceptable basis to document the code.

Thanks! Will try review it today or tomorrow.

PR #12, opened 2024-03-29
Added doc-comments to tensor.rs

nestordemeure:
Following the discussion in [issue11](https://github.com/elodin-sys/elodin/issues/11), here is a PR adding doc comments to `tensor.rs`. This is a first test in introducing AI-generated comments to document the library faster, making it easier to pick up by new users.

Here is a preview of what it looks like in conditions (`cargo doc --open --no-deps --all-features`):

image
image

This is a test to think about the viability of this approach. Do feel free to comment and criticize (things can definitely be streamlined by hand, this PR had no manual tweak of the comments so far)!

For archival purposes, here is the prompt used to generate the comments (with GPT-4):

```
You will be given code from `nox`, a Rust implementation of JAX. Your goal is to doc-comment the code:
* adding a top-level doc comment (describing the current file's function),
* doc comment the functions (helping users understand the arguments and outputs of the function),
* doc comment types and traits.

Your output should be only the doc comments and headers, no need to include the rest of the code.

Also, keep your comments short and useful. No point in stating the obvious or writing a long sentence where a short one would do. Those comments should be there to help a new user understand how to use the library from a quick look.
```

### DIFF
--- a/libs/nox/src/tensor.rs
+++ b/libs/nox/src/tensor.rs
@@ -1,3 +1,4 @@
+//! Provides the core functionality for manipulating tensors.
 use crate::{
     AsBuffer, Buffer, Field, FromOp, IntoOp, Noxpr, NoxprScalarExt, Op, Param, Scalar, Vector,
 };
@@ -10,6 +11,8 @@ use std::{
 };
 use xla::{ArrayElement, ElementType, NativeType};
 
+/// Represents a tensor with a specific type `T`, dimensionality `D`, and parameterization `P`.
+/// `P` dictates the underlying representation and operations available on the tensor.
 #[repr(transparent)]
 pub struct Tensor<T, D: TensorDim, P: Param = Op> {
     pub(crate) inner: P::Inner,
@@ -27,12 +30,18 @@ where
     }
 }
 
+/// Trait for items that can be represented as tensors.
+/// Specifies the type of the item, its tensor representation, and dimensionality.
 pub trait TensorItem {
+    /// The type of the item that can be converted from an operation.
     type Item: FromOp;
+    /// The tensor type associated with the item.
     type Tensor<D>
     where
         D: TensorDim;
+    /// The dimensionality of the tensor.
     type Dim: TensorDim;
+    /// The element type of the tensor.
     const ELEM: ElementType;
 }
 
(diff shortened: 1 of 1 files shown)


libs/nox/src/tensor.rs
```diff
@@ -389,10 +406,16 @@ impl<T, D: TensorDim> AsBuffer for Tensor<T, D, Buffer> {
     }
 }
 
+/// Trait for mapping dimensions in tensor operations.
+/// Allows for transforming and replacing dimensions in tensor types.
 pub trait MapDim<D> {
+    /// The item associated with the dimension mapping.
```
akhilles: We probably don't need these inner comments.

libs/nox/src/tensor.rs
```diff
@@ -495,7 +520,9 @@ impl<const N: usize> DefaultMap for Const<N> {
     type DefaultMapDim = Mapped;
 }
 
+/// Trait representing the concatenation of dimensions.
 pub trait DimConcat<A, B> {
+    /// The resulting dimension type after concatenation.
```
akhilles: This is implied.

libs/nox/src/tensor.rs
```diff
@@ -640,9 +671,12 @@ impl<T: TensorItem, D: TensorDim> Tensor<T, D> {
     }
 }
 
+/// Trait for indexing into tensors, allowing for the extraction of sub-tensors or elements based on indices.
 pub trait TensorIndex<T, D: TensorDim> {
+    /// The output type resulting from the indexing operation.
```
akhilles: This is implied.

akhilles (commented):
This is pretty good! The top level comments are all fine. Some of the comments on associated types and trait functions seem a bit unnecessary. Especially for `Out`, `Output`. @sphw, could you take a look as well?

akhilles:
The commit message has a typo btw

nestordemeure:
Thanks, I have fixed the commit text and removed inner comments.

libs/nox/src/tensor.rs
```diff
@@ -10,6 +11,8 @@ use std::{
 };
 use xla::{ArrayElement, ElementType, NativeType};
 
+/// Represents a tensor with a specific type `T`, dimensionality `D`, and parameterization `P`.
```
akhilles: "underlying representation `P`" is a bit more accurate.

libs/nox/src/tensor.rs
```diff
@@ -10,6 +11,8 @@ use std::{
...
+/// `P` dictates the underlying representation and operations available on the tensor.
```
akhilles: "operations available on the tensor" is not quite accurate. Would just drop this sentence, and change the previous sentence to include the "underlying representation" part.

akhilles (commented):
Just have a nitpick on the `Tensor` comment, but ok to merge once that's fixed.

libs/nox/src/tensor.rs
nestordemeure: Done!

akhilles (approved):

PR #23, opened 2025-04-17
Add support for local user modules

akhilles:
See README: https://github.com/elodin-sys/elodin/tree/aleph-dev-nix/images/aleph#add-user-for-local-development.


sphw (approved):

.clang-format
```diff
@@ -1,2 +0,0 @@
-BasedOnStyle: WebKit
-Standard: "c++20"
```
sphw: Can you add these back? They are used for the C++ code in the examples, otherwise editors will try and reformat in a different style

akhilles: added back, looks like only the .clangd file was causing this issue with C code:
image

PR #40, opened 2025-04-28
feat(os): add flake template

sphw:


akhilles (approved):
as a followup, would suggest adding some readme instructions of how to do `nix flake init`.

PR #56, opened 2025-05-08
Add VTableBuilder to C++ API

sphw:
These changes let you fully specify a VTable in pure C++ without the Lua API. It is designed to look very similar to the Rust and Lua APIs.

```cpp
auto table = builder::vtable({
    builder::field<Foo, &Foo::time>(builder::schema(PrimType::F64(), {}, builder::pair(1, "time"))),
    builder::field<Foo, &Foo::mag>(builder::schema(PrimType::F64(), {}, builder::pair(1, "mag")))
});
auto vtable_msg = VTableMsg {
    .id = {1,0},
    .vtable = table,
};
```


libs/db/cpp/db.hpp
```diff
@@ -0,0 +1,3728 @@
...
+      // Update the input span to point past the decoded data
+      input = input.subspan(slice.len);
+    }
+    return result;
+  }
+
+  postcard_error_t decode_raw(postcard_slice_t* slice) {
+    postcard_error_t result;
+    uint8_t val0;
+                        result = postcard_encode_u8(slice, val0);uint8_t val1;
```
akhilles: The formatting here looks odd.

libs/db/cpp/db.hpp
```diff
@@ -0,0 +1,3728 @@
...
+#include <vector>
+#include <optional>
+
+
+#if __has_include("postcard.h")
+#include "postcard.h"
+#endif
+
+
+#ifndef POSTCARD_H
```
akhilles: Could we avoid duplicating the contents of `postcard.h`? It's a bit annoying to maintain. I think requiring end users to import 2 header files is fine.

sphw: This is an out-generated file that automatically imports postcard.h. So you just run `elodin-db gen-cpp` and it will auto generate

libs/db/cpp/db.hpp
sphw: I'll try and fix this. Its because it is autogenerated code, so the formatting is kind hard to get right

libs/db/examples/db.hpp
```diff
@@ -1993,4 +1978,2100 @@ struct MsgStream {
...
+  }
+
+  postcard_error_t decode_raw(postcard_slice_t* slice) {
+    postcard_error_t result;
+    result = postcard_decode_u16(slice, &arg);
+    if(result != POSTCARD_SUCCESS) return result;
+
+    uint8_t val0;
+                        result = postcard_encode_u8(slice, val0);uint8_t val1;
+                        result = postcard_encode_u8(slice, val1);id =  std::tuple<uint8_t, uint8_t>(val0, val1);
```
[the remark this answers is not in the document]
sphw: fixed

libs/db/examples/client.cpp
```diff
@@ -94,28 +97,54 @@ try {
...
-    auto msg_stream_msg_buf_span = std::span<uint8_t>(reinterpret_cast<uint8_t*>(msg_stream_msg_buf.data()), msg_stream_msg_buf.size());
-    // std::println("msg stream encoded {}", msg_stream_msg_buf_span);
+    // std::println("msg stream encoded {}", msg_stream_msg_buf);
 
     sock.write_all(msg_stream_msg_buf.data(), msg_stream_msg_buf.size());
 
+    // sends vtable for sensor data
+
+    auto table = builder::vtable({
+        builder::field<SensorData, &SensorData::time>(builder::schema(PrimType::I64(), {}, builder::pair(1, "time"))),
```
akhilles: update entity id

akhilles (approved):

PR #71, opened 2025-05-24
Replace sidebars with tiles

sphw:
https://github.com/user-attachments/assets/8e7d2235-138c-4ca8-9b30-34dc18f23ba7


akhilles (commented):
Overall love this change, but I think there should be some faster / easier way to bring up the standard setup of entity list + component inspector panels, in whatever layout. Just to avoid regressing in the open editor -> show useful information flow.

Or maybe we fully commit to presets? Add the ability to reference them from the cli / launcher.

akhilles:
can prolly get rid of these buttons:

image

sphw:
Yeah I agree that it needs a little more ergonomics. I think having a palette item that spawns the side-panels automatically would help. I'd also like a way to pin the panels so they always stay on the side. Since this change is relatively small and self-contained I will hold off on merging till I fix those

PR #72, opened 2025-05-27
feat(fw): add ADC monitoring and voltage/current/temperature sensing

akhilles:
- Add ADC driver with 16-bit precision and VREFINT calibration
- Add STM32H7 internal temperature sensor with factory cal points
- Add RTC backup battery (VBAT) voltage monitoring
- Add voltage/current monitor for VIN, main battery, and aux power
- Configure PLL2 at 25MHz for optimal 16-bit ADC performance
- Integrate monitoring data into blackbox logging and debug output


fsw/sensor-fw/src/adc.rs
```diff
@@ -0,0 +1,309 @@
...
+const TS_CAL1_TEMP: f32 = 30.0;
+const TS_CAL2_TEMP: f32 = 110.0;
+
+/// Safe interface for reading factory calibration values from memory
+///
+/// # Safety
+/// This function reads from well-defined factory calibration memory addresses
+/// that are guaranteed to be valid on STM32H7 microcontrollers. These addresses
+/// contain factory-programmed calibration data and are safe to read.
+fn read_factory_u16(addr: usize) -> u16 {
```
sphw: I think this function shouldn't be considered safe because it could be used in an unsafe way,. In the code in its current state it is safe, but you usually mark a function unsafe if it can be potentially used in an unsafe way. I would also make it take the address as a const ptr, then in the const defs you can cast the usizes to the ptrs.

sphw (approved):
Should we try and upstream the ADC support in the STM32HAL?

fsw/sensor-fw/src/adc.rs
akhilles: fixed

PR #74, opened 2025-05-28
Add support for SQL plots

sphw:
This PR adds a new type of plot called an SQL plot that lets the user plot arbitrary SQL queries. To make this change manageable I refactored a large portion of the existing plotting code. The main changes were simplifying the zooming and panning math – lots of it unnecessary once I actually wrote it out by hand and did the algebra. A number of tasks were moved out of the Plot widget into systems, this allows those code pieces to be re-used for other plot types. The automatic bound-setting code was also refactored to use a much simpler code path. These changes will hopefully make the plot code much easier to follow, there are still a few places that are confusing, but they were much harder to refactor into a nicer place.

This PR also adds two functions called `fft` and `fftfreq` that allow you to perform FFTs of time series data

https://github.com/user-attachments/assets/d276a7b6-cf9a-4f68-ad1c-fe574eea8d27


libs/nox-py/examples/cube-sat.py
```diff
@@ -622,29 +622,31 @@ class Debug(el.Archetype):
...
 
 w.spawn(
-    el.Panel.vsplit(
-        el.Panel.hsplit(
-            el.Panel.viewport(
-                track_entity=sat,
-                track_rotation=False,
-                pos=[7.0, 0.0, 0.0],
-                looking_at=[0.0, 0.0, 0.0],
+    el.Panel.sidebars(
```
akhilles: could you update the examples in the top-level examples dir as well?

akhilles (approved):

PR #79, opened 2025-06-07
feat(fw): initialize USB hub over smbus

akhilles:
This is needed for the USB hub to function. In the next revision, instead of writing directly to the hub, this'll probably write to an i2c eeprom instead. Turns out there's a lot of annoying stuff to configure on a hub for it to work well in all scenarios and strapping resistors aren't robust enough.

### DIFF
--- a/fsw/sensor-fw/src/fm24cl16b.rs
+++ b/fsw/sensor-fw/src/fm24cl16b.rs
@@ -1,6 +1,5 @@
-use crate::i2c_dma;
-use core::ops::DerefMut;
-use embedded_hal::i2c::I2c as I2cTrait;
+use crate::i2c_dma::I2cRegs;
+use embedded_hal::i2c::{I2c as I2cTrait, Operation};
 use hal::i2c;
 
 const BASE_ADDRESS: u8 = 0x50; // Base address for FM24CL16B
@@ -9,7 +8,7 @@ const PAGE_SIZE: usize = 256; // 256 bytes per page
 
 #[derive(Debug, defmt::Format)]
 pub enum Error {
-    I2cDma(i2c_dma::Error),
+    I2c(i2c::Error),
     CommunicationFailed,
     AddressOutOfRange,
     InvalidLength,
@@ -17,13 +16,7 @@ pub enum Error {
 
 impl From<i2c::Error> for Error {
     fn from(err: i2c::Error) -> Self {
-        Error::I2cDma(i2c_dma::Error::I2c(err))
-    }
-}
-
-impl From<i2c_dma::Error> for Error {
-    fn from(err: i2c_dma::Error) -> Self {
-        Error::I2cDma(err)
+        Error::I2c(err)
     }
 }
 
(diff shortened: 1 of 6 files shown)


fsw/sensor-fw/src/fm24cl16b.rs
```diff
@@ -74,29 +74,26 @@ impl Fm24cl16b {
...
             self.set_page(page);
         }
 
-        // Set word address then read byte using direct I2C calls
-        I2cTrait::write(i2c_dma.deref_mut(), self.i2c_address, &[word_address])?;
+        // Set word address then read byte using I2C transaction
+        let word_addr_buf = [word_address];
         let mut data = [0u8];
-        I2cTrait::read(i2c_dma.deref_mut(), self.i2c_address, &mut data)?;
+        let mut ops = [Operation::Write(&word_addr_buf), Operation::Read(&mut data)];
```
sphw, line 88: Why move off DMA? It probably doesn't matter that much for this method since it is doing one byte at a time – was there some issue?

fsw/sensor-fw/src/usb2513b.rs
```diff
@@ -0,0 +1,281 @@
...
+            }
+        }
+    }
+
+    /// Read and dump the current configuration registers for debugging
+    pub fn dump_config(&self, i2c_dev: &mut i2c::I2c<I2cRegs>) -> Result<(), Error> {
+        defmt::info!("USB HUB: Reading current configuration...");
+
+        // We access registers up to 0x10 (power_on_time), so need at least 0x10 + 1 bytes + 1 for byte count
+        const MIN_REQUIRED_BYTES: usize = 0x10 + 1 + 1;
```
sphw, line 163: Kinda a nitpick, but I think this should be a MAX_REQUIRED_BYTES right? Because we create a buffer with this constant

fsw/sensor-fw/src/usb2513b.rs
```diff
@@ -0,0 +1,281 @@
...
+
+        // Raw hex dump first
+        defmt::info!("USB HUB: Raw config: {:02X}", config_buffer);
+
+        // Skip the first byte (byte count) and use actual register data
+        let byte_count = config_buffer[0];
+        defmt::info!("USB HUB: Byte count: {}", byte_count);
+
+        // Validate we have enough data for all parsing operations
+        if config_buffer.len() < MIN_REQUIRED_BYTES {
```
sphw: I think this condition will always pass right? If config_buffer is initialized with`MIN_REQUIRED_BYTES` then this check will always fall through. It might make sense to check `byte_count` instead to ensure that you received enough data from the dump operation

fsw/sensor-fw/src/usb2513b.rs
```diff
@@ -0,0 +1,281 @@
...
+            usb_ids.pid,
+            usb_ids.did
+        );
+
+        // Parse config bytes with proper bitfields
+        let cfg1 = ConfigByte1::from(regs[6]);
+        defmt::info!(
+            "USB HUB: Config1(0x{:02X}): self_powered={}, hs_disable={}, multi_tt={}, eop_disable={}",
+            regs[6],
+            cfg1.self_powered(),
```
sphw, line 213: Is there a reason to both derive Format on ConfigByte1 and manually print out all of the registers?

fsw/sensor-fw/src/usb2513b.rs
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        // Configure with standard self-powered config
+        defmt::info!("USB HUB: Writing configuration...");
+        defmt::info!("USB HUB: Config payload: {:02X}", SELF_POWERED_CONFIG);
+        let mut ops = [Operation::Write(&SELF_POWERED_CONFIG)];
+        i2c_dev.transaction(self.address, &mut ops)?;
+        defmt::info!("USB HUB: Configuration written successfully");
+
+        // Finalize with USB_ATTACH
```
sphw, line 268: What does this step do? Does it emulate a real usb attach event for the hub to force enumeration?

fsw/sensor-fw/src/fm24cl16b.rs
akhilles: I wasn't using DMA previously either. I was dereferencing the internal i2c struct, and it was easier to just share that directly. DMA would work fine, but its not as useful when all the ops are blocking.

fsw/sensor-fw/src/usb2513b.rs
akhilles: The min is a bit more appropriate when thinking in terms of the smbus block read layer on top of i2c. In smbus, the response to an i2c read is prefixed with # of bytes, which can be less than the # of bytes you requested. And you have to check that enough bytes are sent.

fsw/sensor-fw/src/usb2513b.rs
akhilles: yeah I meant to check `byte_count`

fsw/sensor-fw/src/usb2513b.rs
akhilles: I tend to just add `defmt::Format` to structs by default, but its not super useful for `bitfield_struct::bitfield` structs unfortunately. It just prints the decimal value of the backing integer:
```
ConfigByte1(156)
```

fsw/sensor-fw/src/usb2513b.rs
akhilles: It does signal a USB attach event to the upstream host, but it's also a way to commit the hub config. As far as I can tell, there's no other way to indicate that I'm done configuring the hub

PR #81, opened 2025-06-07
feat(fw): improve openocd workflow

akhilles:
probe-rs has been frustratingly buggy over multiple versions. I've found the openocd workflow to be much more reliable.

### DIFF
--- a/fsw/openocd/aleph.cfg
+++ b/fsw/openocd/aleph.cfg
@@ -1,5 +1,7 @@
 source [find interface/cmsis-dap.cfg]
 transport select swd
+adapter speed 12000
 set CHIPNAME stm32h747xit6
 source [find target/stm32h7x_dual_bank.cfg]
-reset_config srst_only srst_nogate
+reset_config none
+$_CHIPNAME.cpu0 cortex_m reset_config sysresetreq
--- a/fsw/openocd/flash.sh
+++ b/fsw/openocd/flash.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+IFS=$'\n\t'
+
+# Parse arguments
+USE_DEFMT=false
+ELF=""
+
+while [[ $# -gt 0 ]]; do
+    case $1 in
+        --defmt)
+            USE_DEFMT=true
+            shift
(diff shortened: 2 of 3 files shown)


x46085:
Validated:

image

Note the panic at the end is likely because I'm using a V1 board with V2-updated FSW

akhilles:
> What's the latest probe-rs bug that led to making this change?

the defmt rtt stream just dies after ~5s with:
```
ERROR probe_rs::architecture::arm::core::armv7m: The core is in locked up status as a result of an unrecoverable exception
Error: Error communicating with the probe.
```

but attaching to the rtt stream with defmt-print works fine. also the probe-rs commands other than `run` have regressed quite a bit:
```
akhilles@evil ~/s/e/f/sensor-fw (aleph.usb-debug) [1]> probe-rs info
Probing target via JTAG
-----------------------

Error while probing target: The protocol 'JTAG' could not be selected.

Caused by:
    The probe does not support the JTAG protocol.
Probing target via SWD
----------------------

ERROR probe_rs::architecture::arm::memory::romtable:    Failed to read component information at 0x0.
 WARN probe_rs::architecture::arm::memory::romtable: Component at 0xe0001000: CIDR0 has invalid preamble (expected 0xd, got 0x0)
 WARN probe_rs::architecture::arm::memory::romtable: Component at 0xe0001000: CIDR2 has invalid preamble (expected 0x5, got 0x0)
 WARN probe_rs::architecture::arm::memory::romtable: Component at 0xe0001000: CIDR3 has invalid preamble (expected 0xb1, got 0x0)
 WARN probe_rs::architecture::arm::memory::romtable: Component at 0xe0000000: CIDR0 has invalid preamble (expected 0xd, got 0xb1)
 WARN probe_rs::architecture::arm::memory::romtable: Component at 0xe0000000: CIDR1 has invalid preamble (expected 0x0, got 0x1)
 WARN probe_rs::architecture::arm::memory::romtable: Component at 0xe0000000: CIDR2 has invalid preamble (expected 0x5, got 0xb1)
```

every release seems to fix 1 thing, but then has 4 other regressions. I just don't have much confidence in the project anymore

PR #82, opened 2025-06-07
Add a C blinky

akhilles:
This is a minimal, fully understandable reference for board power and clock setup. It's ~240 lines in a single C file with zero dependencies - just direct register manipulation that anyone can read and debug. When hardware issues come up, having a baseline that definitely works helps isolate problems quickly. It also includes a full RTT target implementation.

### DIFF
--- a/fsw/c-blinky/main.c
+++ b/fsw/c-blinky/main.c
@@ -1,4 +1,5 @@
 typedef unsigned int U32;
+typedef unsigned long long U64;
 
 // Base addresses
 #define RCC_BASE     0x58024400UL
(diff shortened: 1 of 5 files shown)


fsw/c-blinky/main.c
```diff
@@ -0,0 +1,238 @@
...
+    while (dest < &_edata) *dest++ = *src++;
+    dest = &_sbss;
+    while (dest < &_ebss) *dest++ = 0;
+
+    main();
+    while (1);
+}
+
+void Default_Handler(void) {
+    while (1);
```
sphw: Do you actually want to loop forever as the default behavior for all interrupts? I think generally you would want to trigger a breakpoint so the debugger could figure out what happened? Or clear the interrupt register and drop back to the previous state

fsw/c-blinky/main.c
```diff
@@ -0,0 +1,238 @@
+typedef unsigned int U32;
```
sphw, line 1: This is mainly a stylistic thing, but is there any reason to not name this uint32_t. That way it matches the C99 type names

fsw/c-blinky/main.c
```diff
@@ -0,0 +1,238 @@
...
+#define PLL1_N             100
+#define PLL1_P             2
+#define PLL1_Q             8
+#define RTT_BUFFER_SIZE_UP 1024
+
+// RTT structures
+typedef struct {
+    const char*       sName;
+    char*             pBuffer;
+    unsigned          SizeOfBuffer;
```
sphw, line 102: This might be a C style thing I'm not familiar with, but some of these field names are in PascalCase and others are in camelCase

akhilles: I just used the same style as segger: https://github.com/SEGGERMicro/RTT/blob/ff57c3d516ea59b0a2d4d5cafb33a48572b09c0b/RTT/SEGGER_RTT.h#L304. Not sure why they did this.

fsw/c-blinky/main.c
akhilles: for `uint32_t`, you need to bring in `stdint.h` and it's a bit of a hassle to ensure the right headers are available. and I wouldn't want to conflict with `stdint.h` if someone does pull it in.

akhilles:
> What's the motivation for doing this in C? I get the no-dependency idea, but you could likely do something similar in Rust. In Rust you would probably want 3 dependencies `cortex_m_rt`, `rtt_target`, and maybe the stm32h7 PAC. But that would let you skip all the manual register definitions, but not use a HAL

The main goal with this blinky was to show power + clock tree setup for the board, and I think the simplest way to do that was with the `system_init` and `clock_config` C functions. There's a very direct relationship from those functions to RM0399. With the Rust pac, you still have the indirection + somewhat deep call stack and it's not obvious what code like `pac::Peripherals::take()` does. Also, I do think having 0 deps is very valuable here. It's super low maintenance, barely any code, no constant churn from upgrading dependencies. And having a single minimal C example is super useful for porting stuff like Betaflight, PX4. IMO, we should have a Rust blinky as well.

fsw/c-blinky/main.c
```diff
@@ -0,0 +1,259 @@
...
+    RCC_AHB4ENR |= RCC_AHB4ENR_GPIOEEN;
+    // Configure PE5 (red LED) as output
+    GPIOE_MODER &= ~(3 << (RED_LED_PIN * 2));
+    GPIOE_MODER |= (1 << (RED_LED_PIN * 2));
+    GPIOE_OTYPER &= ~(1 << RED_LED_PIN);
+    GPIOE_OSPEEDR &= ~(3 << (RED_LED_PIN * 2));
+}
+
+static void systick_init(void) {
+    SYSTICK_RVR = 400000 - 1;
```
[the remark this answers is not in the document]
akhilles: the m7 core uses SYSCLK, the m4 core uses HCLK

PR #88, opened 2025-06-20
Entities no more

sphw:
Broadly does what it says on the tin, removes entities as a concept. 

## Why

Initially elodin-db, and the rest of the systems were designed around ECS as a data model. The idea was that the sim, and flight software would all use ECS. In practice flight software was never developed using ECS, because creating a safety-critical ECS system was difficult. Plus it's not clear that flight software is actually simplified by the ECS data model. The simulation side did use ECS, but even that proved to be a sometimes awkward designed pattern.

The thing that did work really well was treating the sim and flight software as a collection of tagged components with a pair of (entity_id, component_id). Entity id was always kinda the odd-man out though. It was nominally meant to be allocated by the user manually, unlike component id. It also wasn't clear how to separate things into different entities. Should each motor be its own entity? Should there be a single entity for the entire satellite?

## The new approach
A better approach is nesting. Logically you want something like `drone.motors.0.rpm` and `drone.imu.accel`. We have kept component ids, and they are still just hashes of the component names. The editor is now aware of paths, and uses them to organize components in the UI.

One problem here is how you display a 3D object, which needs a mesh, material, and position. In the past synced a elodin-db entity directly into a Bevy Entity. Now we use a different technique. 3d Objects are a purely client side construct, similar to a plot. The user is free to spawn a 3d object referring to whatever position, mesh, and shape they wish. For backwards compatibility reasons any component using the old asset and world_pos approach will automatically get a 3d object spawned

## What's left

The component list (nee entity list) does not display nested components as a tree. Instead it simple shows you the top level components as a list. A follow up PR will refactor this into a tree

Some work needs to be done to expose 3d Objects, and the like in the editor. This will require some refactoring of the very messy data structures that the viewports, 3d objects, and plots use.


akhilles (commented):
Could you update the db example as well? It's still using `set_entity_name`. And maybe regenerate `db.hpp`.

sphw:
> Could you update the db example as well? It's still using `set_entity_name`. And maybe regenerate `db.hpp`.

I did in the schematic follow up PR. That is pretty much ready to land. So we could just merge them back to back

akhilles (approved):

PR #89, opened 2025-06-24
Add schematics

sphw:


libs/db/examples/client.cpp
```diff
@@ -131,25 +133,25 @@ try {
...
-    sock.send(set_component_name("pressure"));
-    sock.send(set_component_name("humidity"));
-    sock.send(set_entity_name(1, "vehicle"));
+    sock.send(set_component_name("vehicle.imu.mag"));
+    sock.send(set_component_name("vehicle.imu.gyro"));
+    sock.send(set_component_name("vehicle.imu.accel"));
+    sock.send(set_component_name("vehicle.temp.temp"));
+    sock.send(set_component_name("vehicle.temp.pressure"));
+    sock.send(set_component_name("vehicle.temp.humidity"));
+    sock.send(set_component_name("vehicle"));
```
akhilles: Is this needed?

libs/nox-py/src/well_known/gui.rs
akhilles, line 1: Can we replace the python panel config stuff + assets with kdl file? Maybe not in this PR but at least before a release.

libs/db/examples/client.cpp
sphw: It was at one point, but I think I removed the need for it. I'll remove

libs/nox-py/src/well_known/gui.rs
sphw: I'll add in a follow up PR

akhilles (approved):

PR #113, opened 2025-09-09
Use `Self` where possible

adamnemecek:


shanecelis:
Thank you for your PR. I can see the sense in this change, but I feel like it's strictly speaking it's less informative than the original. As a reader I now have to refer the opening brace to see what enumeration I'm in to fully understand the line, so I'm going to decline to merge this change. Thank you for your input. I hope you'll consider submitting other PRs in the future.

PR #117, opened 2025-09-11
Expose Schematic Saving and Loading in Editor

shanecelis:
## Saving
From the editor you can now save a schematic with "Command-P Save Schematic" and it will write it to a `DEFAULT_SCHEMATIC_FILENAME` or prompt for a filename if not present. 

```python
# world.schematic([DEFAULT_SCHEMATIC], [DEFAULT_SCHEMATIC_FILENAME])
world.schematic("""
    graph "rocket.world_pos[4]"
""", "rocket.kdl")
```

## Loading
One can then load the schematic with "Command-P Load Schematic". It will show schematics (.kdl) files present in the working directory and an option to use a file dialog to load from other directories.

### At Startup

If a `DEFAULT_SCHEMATIC_FILENAME` is given, that file will be loaded at startup if it exists. Otherwise, the `DEFAULT_SCHEMATIC` from the code is used.

## IDs

This change also adds an optional `id` argument to spawn: `world.spawn([components], [name], [id])`. Name can be any string. If no ID is given and a name is given, then name is converted to snake_case to use as an ID. ID must be snake_case and will report an error if it's not. Name is currently forgotten after this. Perhaps we can add it back as a human-readable thing later.

This restores the way it used to work in code partly:
```python
css_5 = w.spawn(
    SunSensor(
        value=0.0,
        fov=np.radians(90),
        normal=np.array([-1.0, 0.0, 0.0]),
    ),
    name="Course Sun Sensor 5",
    id="css_5", # This is new. If it weren't here, "course_sun_sensor_5" would be used based on name.
)
```

### DIFF
--- a/libs/nox/src/noxpr/batch.rs
+++ b/libs/nox/src/noxpr/batch.rs
@@ -102,6 +102,7 @@ impl BatchTracer {
         self.visit_inner(expr)
     }
 
+    #[inline]
     /// Visits a `Noxpr` and applies any batch-specific transformations.
     pub fn visit_inner(&mut self, expr: &Noxpr) -> Result<BatchedExpr, Error> {
         let id = expr.id();


libs/elodin-editor/src/ui/command_palette/palette_items.rs
```diff
@@ -1043,7 +1073,8 @@ impl Default for PalettePage {
             create_sidebars(),
             create_3d_object(),
             save_schematic(),
-            save_schematic_db(),
```
x46085, line 1046: What's the plan here? What does this imply?

libs/impeller2/kdl/src/ser.rs
```diff
@@ -118,6 +118,7 @@ fn serialize_graph<T>(graph: &Graph<T>) -> KdlNode {
     let mut node = KdlNode::new("graph");
 
     // Add the EQL query as the first unnamed entry
+    // tracing::warn!("GRAPH EQL {:?}", &graph.eql); this shows the Rocket.Rocket.world...
```
x46085: Intentionally left in?

libs/impeller2/kdl/src/ser.rs
```diff
@@ -730,12 +732,39 @@ object_3d "a.world_pos" {
...
+        //viewport active=#true hdr=#true fov=45.0 show_grid=#false
+        assert_eq!(
+            r#"tabs {
+    viewport active=#true hdr=#true
+    graph a.world_pos name="a world_pos"
+}
+object_3d a.world_pos {
+    sphere radius=0.20000000298023224 r=1.0 g=1.0 b=1.0
+}
+"#,
```
x46085: Would be nice if formatted to be human readable, is that doable or causes issues?

libs/impeller2/kdl/src/ser.rs
```diff
@@ -730,12 +732,39 @@ object_3d "a.world_pos" {
...
     }
 
+    #[test]
+    fn test_roundtrip_rocket_example() {
+        let original_kdl = r#"graph "rocket.fin_deflect[0]" name=Fin "#;
+        let parsed = parse_schematic(original_kdl).unwrap();
+        let serialized = serialize_schematic(&parsed);
+        assert_eq!(
+            r#"graph "rocket.fin_deflect[0]" name=Fin
+"#,
```
x46085: can be one line, or no?

libs/impeller2/wkt/src/path.rs
```diff
@@ -69,13 +69,20 @@ impl ComponentPath {
...
+        // it's actually stored ["a", "a.b", "a.b.c"], which seems strange.
+        // TODO: Consider changing it to store ["a", "b", "c"].
+
+        // for (i, part) in self.path.iter().enumerate() {
+        //     if i > 0 {
+        //         write!(f, ".{}", part.name)?;
+        //     } else {
+        //         write!(f, "{}", part.name)?;
+        //     }
+        // }
```
x46085, line 85: I believe it was intentional, displaying each entry as members of a tree / graph (like in Unreal/Unity). Fine to change to something you find more obvious / easier to understand.

libs/impeller2/kdl/src/ser.rs
shanecelis: It is doable. We just have to choose our precision. How many digits after the decimal point?

libs/impeller2/kdl/src/ser.rs
shanecelis: Nope. Will remove.

libs/nox-py/examples/cube-sat-pysim.py
```diff
@@ -622,55 +618,6 @@ class Debug(el.Archetype):
...
-    object_3d earth.world_pos {
-        glb path="https://storage.googleapis.com/elodin-assets/earth.glb"
-    }
-
-    object_3d sat.world_pos {
-        glb path="https://storage.googleapis.com/elodin-assets/oresat-low.glb"
-    }
-    line_3d sat.world_pos line_width=10.0 perspective=#false
-""")
-
```
x46085, line 673: Is this removed because there is not use of the editor in this example?

libs/elodin-editor/src/ui/command_palette/palette_items.rs
shanecelis: Expose at a later date if needed.

x46085 (approved):

libs/impeller2/kdl/src/ser.rs
x46085: No I meant the tab / spacing, the numbers are fine.

libs/elodin-editor/src/ui/command_palette/palette_items.rs
shanecelis: Latest looks like this:

```rust
            // TODO: Don't enable until we have a clear save and load story for db.
            //save_schematic_db(),
```

libs/nox-py/examples/cube-sat-pysim.py
shanecelis: Yes. It doesn't do anything when not used with the editor.

libs/impeller2/wkt/src/path.rs
shanecelis: I'm hesitant to touch it because it looks like it might be used as keys. You may have a tree
```
  a
 / \
b   c
   /
  b
```
You may want the 'b's to have different hashes. That'd be the reason to do it like this.

libs/impeller2/kdl/src/ser.rs
shanecelis: I was going to say no but yes, we can trim it (without allocating another string too).

libs/impeller2/kdl/src/ser.rs
shanecelis: [Chats.] Ok, done.

PR #119, opened 2025-09-11
fix(aleph): macos networking on aleph

labrnth:
Building off of https://github.com/elodin-sys/elodin/pull/95 which broke aleph ethernet networking for macos.

It turns out that identical host/client mac address break NDP/ARP resolution on macos . They need to be unique. There is a difference between how Linux and MacOS handle ARP/NDP the outcome is the difference in this apparent behavior. The exact mechanism of action is left as an exercise for the reader :)


x46085:
Nice discovery! I know that was a head scratcher.

PR #130, opened 2025-09-17
Release v0.15.0-alpha.1

shanecelis:
- **chore: Bump to 0.15.0-alpha.1.**
- **doc: Update CHANGELOG.**
- **chore: Run `just public-changelog`.**


docs/public/config.toml
```diff
@@ -44,7 +44,7 @@ github = "https://github.com/elodin-sys"
 twitter = "https://twitter.com/elodinsystems"
 email = "info@elodin.systems"
 ganalytics = ""
-version = "v0.15.0"
+version = "v0.15.0-alpha.1"
```
x46085: Just FYI, this version field drives this UI on the docs site:
image

Which with this change would look like this:
image

Which I don't imagine is what you want

x46085: Should revert this

nix/pkgs/elodin-py.nix
```diff
@@ -62,19 +62,21 @@
 
   wheelName = "elodin";
   wheelSuffix = "cp310-abi3-linux_${arch}";
+  # Convert version format from 0.15.0-alpha.1 to 0.15.0a1 for wheel filename
+  wheelVersion = lib.strings.replaceStrings ["-alpha."] ["a"] version;
```
x46085, line 66: I'm a little worried about unintended consequences of this. No interest in just adopting the "0.15.0a1" convention so it's consistent everywhere? I guess one way to find out if this will work downstream.

shanecelis: I wouldn't want to rob crates.io to pay Pypi and since we only have one Pypi thing, I think it's a tolerable deviation.

docs/public/config.toml
shanecelis: Will do.

shanecelis: Surprisingly the PR bot still shows the issue, but when I ran a local doc web server it's shown corrected. I imagine when merged, it'll be shown correctly.

PR #131, opened 2025-09-18
feat(aleph) - allow custom save location for db

labrnth:
In elodin-db module, the database will now be saved as a _sub-folder_ of `/db.` The previous behavior was to save in `/db` directly. The subfolder name is customizable and can be passed as an argument to the service, where:

> `systemctl start elodin-db@NAME.service`

will result in a database saved to` /db/name`. This is useful for test applications where it is desirable to initialize the database in new locations.

Note: the parent folder (`/db`) is also customizable as a variable in the elodin-db nix module.

### DIFF
--- a/images/aleph/modules/elodin-db.nix
+++ b/images/aleph/modules/elodin-db.nix
@@ -21,21 +21,44 @@ in {
         Whether to automatically open the specified ports in the firewall.
       '';
     };
+    dbFolderName = lib.mkOption {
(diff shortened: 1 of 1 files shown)


x46085 (approved):
Great improvement. This prompts the question on where cool Aleph related features like this get explained. The Aleph README has some basic info, but there's a ton not really illuminated. I'll have to come up with a plan for that.

Issue #139, opened 2025-09-18
Enhanced plot zoom/pan interactions


zxq82lm:
# Description
Ability to zoom/pan multiple plots/graphs simultaneously along the same axis (for example synchronizing the X-axis for time-based data).

# To do List
- Create a small lock on each window. If it’s closed, the current graph inherits the zoom level (usually the x-axis unit, typically time) from another locked window. If there’s only one locked window, nothing happens.
- ...

PR #144, opened 2025-09-18
feat(plot): Enhanced plot zoom/pan interactions

zxq82lm:
See this issue: https://github.com/elodin-sys/elodin/issues/139

# Feature (lock behavior on plots)
- First lock activated
   - The first plot-graph whose lock is toggled becomes the leader.
   - No visible effect at first, but this plot will serve as the reference for `zoom_x`.
- Second lock (and subsequent ones)
   - When another plot-graph is locked, it automatically synchronizes with the `zoom_x` of the already locked group.
   - As a result, all locked plots share the same `zoom_x` state.
- Propagation of `zoom_x`
   - Whenever a user applies a `zoom_x` (via the UI) on a locked plot, this zoom is transmitted to all other locked plots.
   -This works in a peer-to-peer logic: each locked plot can temporarily become the “leader” while it’s being interacted with.

# `zoom_x` in/out
- On macOS: CMD + mouse wheel

# egui_materials-icons
This branch modifies `elodin-editor/Cargo.toml` to include the following:
```toml
[dependencies]
egui_material_icons = "0.3"
```

### DIFF
--- a/libs/elodin-editor/src/ui/plot/widget.rs
+++ b/libs/elodin-editor/src/ui/plot/widget.rs
@@ -387,6 +387,43 @@ impl TimeseriesPlot {
         let response = ui.allocate_rect(self.rect, egui::Sense::click_and_drag());
         let pointer_pos = ui.input(|i| i.pointer.latest_pos());
 
+        // Small clickable lock in the top-right corner of the plot.
+        // It only toggles `graph_state.locked` (no functional effect yet).
+        {
+            let lock_size = egui::vec2(20.0, 20.0);
+            let lock_pos  = egui::pos2(
+                self.inner_rect.max.x - lock_size.x - 6.0,
+                self.rect.min.y + 6.0,
+            );
+
+            egui::Area::new(egui::Id::new(("plot_lock_btn", (graph_state as *const _) as usize)))
+                .order(egui::Order::Foreground)
+                .fixed_pos(lock_pos)
+                .show(ui.ctx(), |ui| {
+                    let (rect, resp) = ui.allocate_exact_size(lock_size, egui::Sense::click());
+
+                    // subtle hover feedback
+                    if resp.hovered() {
+                        ui.painter().rect_filled(
+                            rect.expand(2.0),
+                            egui::CornerRadius::same(6), // u8 with your egui version
+                            get_scheme().bg_secondary.opacity(0.6),
+                        );
+                    }
+
+                    // vector lock icon (defined at bottom)
+                    paint_lock_icon(ui, rect, graph_state.locked);
+
+                    // callback is intentionally "empty": toggle only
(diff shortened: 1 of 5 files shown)


shanecelis:
This is an exciting change!

# Current Observed Behavior

I played around with it a little bit. I can see it working sometimes. It seems like the first locked graph can control the zoom of the others. However, the other locked panels don't seem to be able to control the zoom.

# Desired Behavior

I'd like for all panels that are locked to be "peers", that is, if I change the zoom on any one of them that are locked, it changes them all.

shanecelis:
https://github.com/user-attachments/assets/1e70610f-b984-423c-9a44-845e14817389

I took a little screencast (Command-Option-5 on macOS) to show you what this behavior looks like.

# Observed Behavior

Graphs seem to get stuck and can't translate sometimes (click drag should translate). If I zoom in (with the scroll wheel) sometimes I can't seem to zoom back out. Indeed, I can't even see the labels for the x-axis sometimes. Clicking the lock on or off does reset the zoom, which is a nice side benefit but it makes me wonder if we may want a "max bounds" or "reset" zoom button in the future.

# Desired Behavior

I'd like for zooming and translation to always be available on the graphs.

libs/elodin-editor/src/ui/plot/widget.rs
```diff
@@ -994,21 +1326,14 @@ fn sigfig_round(x: f64, mut digits: i32) -> f64 {
 pub fn pretty_round(num: f64) -> f64 {
     let mut multiplier = 1.0;
     let mut n = num;
-
-    // Handle negative numbers
     let is_negative = n < 0.0;
     n = n.abs();
-
-    // Find the appropriate multiplier for the decimal places
```
shanecelis, line 1002: I'm a little hesitant to lose some of these comments.

libs/elodin-editor/src/ui/plot/widget.rs
```diff
@@ -675,10 +707,10 @@ pub fn sync_bounds(
     let outer_ratio = (rect.size() / inner_rect.size()).as_dvec2();
     let pan_offset = graph_state.pan_offset.as_dvec2() * DVec2::new(-1.0, 1.0);
     PlotBounds::from_lines(&selected_range, earliest_timestamp, y_min, y_max)
-        .zoom_at(outer_ratio, DVec2::new(1.0, 0.5)) // zoom the bounds out so the graph takes up the entire screen
```
shanecelis, line 678: Some of these comments seem helpful too.

shanecelis (approved):
I suggested we retain a few comments. Instead of having you question whether each removed comment should be retained, I simply added a commit that retained ones I thought were valuable.

I exercised this behavior on drones, and it looks and feels great. This is approved with a big thumbs up!

PR #145, opened 2025-09-21
Elodin-DB: Rust Example - Rocket Trim Client Control

x46085:


libs/db/examples/rust_client/README.md
```diff
@@ -0,0 +1,485 @@
...
+   ```python
+   w.run(system, 
+         sim_time_step=1/120.0,
+         run_time_step=1/120.0)  # Real-time execution
+   ```
+
+3. **Client Control**: The Rust client sends control commands:
+   ```rust
+   // Sends sinusoidal trim commands (±10° @ 0.25Hz)
+   let trim_value = amplitude * (2.0 * PI * frequency * elapsed).sin();
```
shanecelis, line 231: This sets a variable. But where's the function that sends it somewhere?

libs/db/examples/rust_client/README.md
```diff
@@ -0,0 +1,485 @@
...
+```
+
+### Command Line Options
+
+```bash
+./target/release/rust_client --help
+
+Options:
+  -H, --host <HOST>    Host address of the Elodin-DB server [default: 127.0.0.1]
+  -p, --port <PORT>    Port of the Elodin-DB server [default: 2240]
```
shanecelis: The elodin-db uses a colon delimited host and port specification, e.g., "[::]:2240". Here we separate them, i.e., -h [::] -p 2240. I don't have a particular preference, but I'd probably want it to be uniform and defer to what came before. Can we support doing "-h [::]:2240" as a concession to the prior standard?

libs/db/examples/rust_client/README.md
```diff
@@ -0,0 +1,485 @@
...
+
+```rust
+// C/C++ approach - manual registration
+send_set_component_metadata(sock, "rocket.mach");
+send_set_component_metadata(sock, "rocket.thrust");
+// ... repeat for each component
+
+// Rust approach - automatic discovery
+let components = discover_components(client).await?;
+// Client now knows about ALL components and their schemas!
```
shanecelis, line 27: This is cool.

libs/nox-ecs/src/impeller2_server.rs
```diff
@@ -181,6 +199,12 @@ pub fn commit_world_head(
                 continue;
             };
             let pair_name = format!("{}.{}", entity_metadata.name, component_metadata.name);
+
+            // Skip writing back external control components
+            if component_metadata.metadata.get("external_control") == Some(&"true".to_string()) {
```
shanecelis: We're allocating a string to do a comparison to satisfy the types. I'd suggest changing this to the following:

```rust
if component_metadata.metadata.get("external_control").map(|s| s == "true").unwrap_or(false) {
```

libs/nox-py/README.md
```diff
@@ -221,6 +240,108 @@ def spring_force(
...
+Run the complete example:
+```bash
+# Terminal 1: Simulation
+python examples/rocket.py run 0.0.0.0:2240
+
+# Terminal 2: External controller
+cargo run --release -p elodin-db-rust-client
+
+# Terminal 3: Visualization
+elodin editor 127.0.0.1:2240
```
shanecelis, line 335: Confirmed that this works.

shanecelis (changes requested):
Request two changes:

1. Permit host flag to accept a port, e.g., "-h localhost:2290".
2. Avoid the string allocation for "external_control" conditional.

Issue #150, opened 2025-09-24
Feature request: Sort elements in create plots UI


shanecelis:
Nat suggested that the list of entities and components ought to be sorted so that they're easier to find.

PR #151, opened 2025-09-24
feature: Sort the components in UI.

shanecelis:
This addresses issue #150: Sorts the components in the UI.

The change is simple. Instead of using a `HashMap` as a backing store for the EQL context, I switched it to a `BTreeMap`, which stores its keys in order. No extra allocations, no on-request sorting required.

I confirmed this by looking at the drones examples and doing a Command-P, "Create Graph", select "drone", and see that all its components are in alphabetical order.

```sh
cd lib/nox-py
cargo run --manifest-path=../../apps/elodin/Cargo.toml editor ../../examples/drone/main.py
```

### DIFF
--- a/libs/elodin-editor/src/lib.rs
+++ b/libs/elodin-editor/src/lib.rs
@@ -981,18 +981,8 @@ pub fn clamp_current_time(
     }
 }
 
-#[derive(Resource)]
+#[derive(Default, Resource)]
 pub struct EqlContext(pub eql::Context);
-
-impl Default for EqlContext {
-    fn default() -> Self {
-        Self(eql::Context::new(
-            HashMap::new(),
-            Timestamp(i64::MIN),
-            Timestamp(i64::MAX),
-        ))
-    }
-}
 
 pub fn update_eql_context(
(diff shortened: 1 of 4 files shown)


libs/elodin-editor/src/lib.rs
```diff
@@ -981,19 +981,9 @@ pub fn clamp_current_time(
...
 
-impl Default for EqlContext {
-    fn default() -> Self {
-        Self(eql::Context::new(
-            HashMap::new(),
-            Timestamp(i64::MIN),
-            Timestamp(i64::MAX),
-        ))
-    }
-}
```
x46085, line 995: Unrelated? Or just unused?

shanecelis: We get it for free with the `#[derive(Default)] EqlContext(Context)`. My guess is `Context` didn't always implement default but now it does so we can take advantage.

zxq82lm (commented):
I ran the indicated test: drone example > Command-P “Create Graph” > select “drone.”
- The proposed list is sorted alphabetically as expected.

PR #153, opened 2025-09-24
Initiate a README for nox/array

zxq82lm:


libs/nox/array/README.md
```diff
@@ -0,0 +1,55 @@
+# nox/array
```
shanecelis: What to call this? Its path is "nox/array". Its package is "nox-array". Its crate is "nox_array". I suggest we call it "nox_array" to match its crate name.

libs/nox/array/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Description
+A small building block of the `nox` crate for representing n‑dimensional arrays (tensors) in a zero‑copy friendly way thanks to `ArrayView`.
+`ArrayView` does not own data: it borrows a buffer (`&[T]` or `&[u8]`) and associates it with a shape (`&[usize]`). This is enough to describe and manipulate multi‑dimensional arrays even in constrained environments (embedded, GPU, WASM), without allocation and without heavy dependencies like ndarray.
+
+## Simple example
+```rust
+use nox_array::ArrayView;
```
shanecelis, line 9: Love the examples. Do the doctests pass as is?

shanecelis (approved):

libs/nox/array/README.md
zxq82lm: I updated the first line to:
```
# nox_array
```

libs/nox/array/README.md
zxq82lm: I actually wasn’t aware of that possibility. Thanks a lot for pointing it out! 🙏
After a few adjustments, running:
```bash
cargo test --doc -p nox-array
```
now gives:
```bash
running 3 tests
test libs/nox/array/src/../README.md - (line 8)  ... ok
test libs/nox/array/src/../README.md - (line 45) ... ok
test libs/nox/array/src/../README.md - (line 30) ... ok
```

PR #156, opened 2025-09-25
Initiate a README for nox_frames

zxq82lm:
- Initiate a README for nox_frames
- All the 3-examples in the README pass their doctests

### DIFF
--- a/libs/nox-frames/src/lib.rs
+++ b/libs/nox-frames/src/lib.rs
@@ -1,3 +1,6 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr( doctest, doc = include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/README.md")))]
+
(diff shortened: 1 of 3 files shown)


libs/nox-frames/src/lib.rs
```diff
@@ -1,3 +1,6 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr( doctest, doc = include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/README.md")))]
```
shanecelis: I would include the readme always instead of conditionally. That way any generated docs will be benefitted by having the README there.

shanecelis:
I pushed a small change that checks the results in the last two doc tests. It is comparing on the floats, which feels brittle. I'd prefer to do a floor() or convert them to i64 or something like that, but I'm actually not sure how you "map" to the scalars within these tensor objects.

PR #159, opened 2025-09-26
Release v0.15.0

shanecelis:
Expose noxpr module for documentation.


shanecelis:
Hmm, I did not mean to close this. It's because I rebased it onto main, which took all its commits away. TIL.

PR #184, opened 2025-10-01
Extend third README example with symbolic graph evaluation

zxq82lm:
I was initially blocked on adding an evaluation for the component-based symbolic graph (third example). @shanecelis  pointed me toward using `noxpr`, and after adding the evaluation, the corresponding doctest now passes successfully.


shanecelis (approved):
Looks good. Tests pass. And it's awesome to have an example of an expression evaluated.

PR #185, opened 2025-10-02
chore (rust) Fix clippy issues

joelreymont:


libs/impeller2/src/types.rs
```diff
@@ -825,7 +828,7 @@ impl<B: IoBuf> MsgBuf<B> {
     }
 
     pub fn try_parse<'a, T: Msg + Deserialize<'a> + 'a>(&'a self) -> Option<Result<T, Error>> {
-        if T::ID == self.id {
+        if T::ID != self.id {
             return None;
         }
         Some(self.parse())
```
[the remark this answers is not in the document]
joelreymont: I claim that Graphite is being dumb here and that the fix is correct.

libs/impeller2/src/types.rs
```diff
@@ -678,24 +678,27 @@ impl LenPacket {
         self.inner[..4].copy_from_slice(&len.to_le_bytes());
     }
 
+    // N.B: repr(C) on Packet prevents us from deriving IntoBytes.
+    // This means we can't use `try_ref_from_bytes` and `try_mut_from_bytes`.
+
     #[inline]
-    pub fn as_packet(&self) -> &Packet {
+    pub fn as_packet_parts(&self) -> (&PacketHeader, &[u8]) {
```
anderspitman: Is this API change necessary? Personally I prefer the old one. I always get confused by tuples (and std::pairs in C++) because I can't remember what the parts are.

How do I reproduce the clippy warnings you're seeing? I poked through the repo and CI stuff but I'm not seeing what we're invoking it with, and when I run just `cargo clippy` in `libs/impeller2` I'm not getting any warnings.

anderspitman (commented):
Mostly looks good, just had one question

libs/impeller2/src/types.rs
joelreymont: Good catch! This won't hit us until I bump crate versions as part of the XLA upgrade. I'll remove it for now.

libs/impeller2/src/types.rs
```diff
@@ -825,7 +825,7 @@ impl<B: IoBuf> MsgBuf<B> {
     }
 
     pub fn try_parse<'a, T: Msg + Deserialize<'a> + 'a>(&'a self) -> Option<Result<T, Error>> {
-        if T::ID == self.id {
```
anderspitman, line 828: Looks like this was completely broken before. Is `try_parse` just never being used?

libs/nox/build.rs
```diff
@@ -48,7 +48,7 @@ fn main() {
     if shared_lib.exists() {
         let shared_lib_name = shared_lib.file_name().unwrap();
         let shared_lib_dest = python_dir.join(shared_lib_name);
-        std::fs::copy(&shared_lib, &shared_lib_dest).unwrap();
```
anderspitman, line 51: Will this silently ignore errors?

libs/impeller2/src/types.rs
joelreymont: I have no idea.

libs/nox/build.rs
joelreymont: It will. It's just a build script, though, so you'll know about it instantly!

zxq82lm:
On my first local build of this PR I checked out the branch and ran the full test suite:
```bash
git fetch origin pull/185/head:pr-185
git checkout pr-185
cargo test --workspace
```
That initial run failed on `tests::test_concurrent_clients` (expected 50, got 49, with a TimeTravel warning).
To investigate further on branch pr-185, I ran the test repeatedly:
```bash
cargo test -p elodin-db-tests --lib -- tests::test_concurrent_clients
for i in {1..20}; do \
  cargo test -q -p elodin-db-tests --lib -- tests::test_concurrent_clients || { echo "FAIL run $i"; break; }; \
done
```
This showed the same intermittent behavior (sometimes 49 instead of 50).
I then switched to the main branch and repeated the same stress test:
```bash
git checkout main
git pull
cargo clean
for i in {1..20}; do \
  cargo test -q -p elodin-db-tests --lib -- tests::test_concurrent_clients || { echo "FAIL run $i"; break; }; \
done
```
I was able to reproduce the failure there as well.
So it seems to be an existing flaky concurrency test, **not something introduced by this PR**. 
This flaky behavior might already be known, but if not, should I open an issue to track it?

joelreymont:
No, it's not something introduced by the PR.

I discovered this issue a while ago and meant to get AI to work on it but haven't had a chance to.

Please create a ticket to track it!

zxq82lm:
See: https://github.com/elodin-sys/elodin/issues/186

zxq82lm (approved):
I ran Clippy (with strict `-D warnings`) and, where relevant, tests on each crate touched by this PR.

### 1. impeller2
```bash
cargo clippy --manifest-path libs/impeller2/Cargo.toml --all-targets -- -D warnings
```
✅ Passed with no warnings.

### 2. nox
```bash
cargo clippy --manifest-path libs/nox/Cargo.toml --lib --no-default-features -- -D warnings
```
✅ Passed with no warnings.

### 3. nox-py
```bash
cargo clippy --manifest-path libs/nox-py/Cargo.toml --lib --no-default-features -- -D warnings || cargo clippy --manifest-path libs/nox-py/Cargo.toml --bins --no-default-features -- -D warnings
```
✅ Passed with no warnings.

### 4. stellarator
```bash
cargo clippy --manifest-path libs/stellarator/Cargo.toml --lib --no-default-features -- -D warnings
```
✅ Passed with no warnings.

### 5. roci-adcs
```bash
cargo clippy --manifest-path libs/roci/adcs/Cargo.toml --lib --no-default-features -- -D warnings
```
✅ Passed with no warnings.

### 6. elodin-db-tests
```bash
cargo clippy -p elodin-db-tests --tests -- -D warnings
cargo test -p elodin-db-tests
```
✅ Clippy clean and tests passed.

Issue #169, opened 2025-09-26
Choose colors in graphs


shanecelis:
# Need

Choose colors in graphs

# Problem

Can't specify colors in graphs

# Solution

Specify colors in KDL. 

```kdl
graph "rocket.fins[2], rocket.fins[3]" {
    color 12.7 9.2 0.5
    color 0.7 2.2 0.5
}
```

Many colors may be specified as children of a graph. Map those colors to each of the lines used within the graph.

# Source

Sung provided this issue.

zxq82lm:
Colors are defined in normalized RGB values within the [0, 1] range. Values greater than 1 are clamped to 1.0, and values less than 0 are clamped to 0.0. To convert from standard 0–255 RGB, simply divide each component by 255.

PR #193, opened 2025-10-03
feat(ui): add custom graph line colors in KDL

zxq82lm:
>Note: The brighter “sun-like” glow you remember in the three-body demo came from setting RGB components well above the old 1.0 limit (e.g., color 10.0 0.0 10.0). That exploited the renderer’s tonemapping to behave almost like an emissive material. In this PR we deliberately removed support for out-of-range floats and now clamp color inputs to the canonical 0–255 per channel. This keeps KDL consistent and avoids ambiguous conversions, but it also means those spheres can no longer over-brighten just by writing color 10.0…. To get the glow back we’ll need a dedicated emissive/high-intensity material path instead of relying on invalid RGB values, so the example currently looks less “sunny” than before.
image
image

## Feature description

The initial request ([#169](https://github.com/elodin-sys/elodin/issues/169)) required RGB colors to be specified as three components in the 0.0–1.0 range in the input file.

- This PR replaces the floating-point RGB format with 0–255 integer values, which are more user-friendly and avoid rounding issues when saving to KDL.
- Floating-point color expressions are ignored to avoid any bias introduced by conversions such as interpreting 1.0 as 1/255: for example `color 1.0 0 0` is ignored.
- It also extends the set of available color definitions with this palette of named values:
```
- black
- blue
- cyan
- gray
- green
- hyperblue
- mint
- orange
- peach
- pink
- pumpkin
- reddish
- slate
- turquoise
- white
- yalk
- yellow
- yolk
```
As a result, there are now two ways to specify a graph color:
```python
graph "rocket.fins[2], rocket.fins[3]" {
    color 255 255 0 # yellow
    color blue
}
```

### 1. With the ball example  
- In `examples/ball/sim.py`, replace the `world.schematic` block with this one below:
```
    world.schematic("""
        hsplit {
            viewport name=Viewport pos="(0,0,0,0, 8,2,4)" look_at="(0,0,0,0, 0,0,3)" hdr=#true show_grid=#true active=#true
        }
        object_3d ball.world_pos {
            sphere radius=0.2 {
                color orange
            }
        }
            color 255 255 0
        }
    """)
```
- Run the ball example
```bash
cargo run --manifest-path=apps/elodin/Cargo.toml editor examples/ball/main.py
```  
- You can test/change colors on the fly in `sim.py` during execution: the balloon’s color hot-reloads instantly in the editor. Invalid color definitions are ignored.
- Save KDL and confirm colors are preserved. Named values are automatically converted to their RGB equivalents. For example: `color green` will be KDL-saved as `color 0 255 0`.

### 2. Unit tests  
Run the parser test to validate color handling:  
```bash
cargo test -p impeller2-kdl test_parse_graph_colors
```

### DIFF
--- a/docs/public/content/reference/migration/to-0.15.md
+++ b/docs/public/content/reference/migration/to-0.15.md
@@ -25,7 +25,7 @@ The ball example had the following code from v0.14.2:
 
 ```python
 ball_mesh = world.insert_asset(el.Mesh.sphere(BALL_RADIUS))
-ball_color = world.insert_asset(el.Material.color(12.7, 9.2, 0.5))
+ball_color = world.insert_asset(el.Material.color(0.1, 0.2, 1.0))
 world.spawn(
     [
         el.Body(world_pos=el.SpatialTransform(linear=jnp.array([0.0, 0.0, 6.0]))),
@@ -38,7 +38,7 @@ world.spawn(
 We can remove the asset handling.
 ```diff
 -ball_mesh = world.insert_asset(el.Mesh.sphere(BALL_RADIUS))
--ball_color = world.insert_asset(el.Material.color(12.7, 9.2, 0.5))
+-ball_color = world.insert_asset(el.Material.color(0.1, 0.2, 1.0))
 world.spawn(
     [
         el.Body(world_pos=el.SpatialTransform(linear=jnp.array([0.0, 0.0, 6.0]))),
@@ -74,7 +74,7 @@ All the preceding information can now be specified via schematics.
         }
         object_3d ball.world_pos {
             sphere radius=0.2 {
-                color 12.7 9.2 0.5
+                color 25 50 255
             }
         }
         line_3d ball.world_pos line_width=2.0 {
@@ -87,15 +87,15 @@ The old code is python, and the new code is specified in a [KDL](https://docs.rs
 
 
 ### `el.Material.color` to `color`
-A color can be specified by its red, green, blue, and optionally its alpha components. 
+A color can be specified by its red, green, blue, and optionally its alpha components as an integer from 0 to 255.
 
 OLD
 ```python
-el.Material.color(12.7, 9.2, 0.5))
+el.Material.color(0.1, 0.2, 1.0))
 ```
 NEW
 ```python
-color 12.7 9.2 0.5 
+color 25 50 255
 ```
 
 It can also be specified by a few names:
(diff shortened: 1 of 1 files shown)


zxq82lm:
@Codex analyze this PR

shanecelis:
This behavior looks good on initial inspection. However, I think we should only accept integers from [0,255] and forget floats. Why? Because I tried to input the following in my three-body.py file:
```
            graph "a.world_pos" name=Graph {
                color 1 0 0
                color 0 1 0
                color 0 0 1
                color 0 0 1
                color 0 0 1
                color 0 0 1
            }
```
And I got a lot of black lines because the '1' is interpreted as 1 out of 255. If I change my '1' to '1.0' then it does what I expected. When we serialize, we serialize it as [0, 255]; so let's only accept integers and not invite confusion. Also integers are better for a human read-able format.

zxq82lm:
> This behavior looks good on initial inspection. However, I think we should only accept integers from [0,255] and forget floats. Why? Because I tried to input the following in my three-body.py file:
> 
> ```
>             graph "a.world_pos" name=Graph {
>                 color 1 0 0
>                 color 0 1 0
>                 color 0 0 1
>                 color 0 0 1
>                 color 0 0 1
>                 color 0 0 1
>             }
> ```
> 
> And I got a lot of black lines because the '1' is interpreted as 1 out of 255. If I change my '1' to '1.0' then it does what I expected. When we serialize, we serialize it as [0, 255]; so let's only accept integers and not invite confusion. Also integers are better for a human read-able format.

Floating-point color expressions are ignored to avoid any bias introduced by conversions such as interpreting 1.0 as 1/255.

shanecelis:
Apologies. I hadn't noticed the visual change this makes. Is it possible for us to permit using HDR values by doing something like this:
```diff
-color 10.0 0.0 10.0
+color 2550 0 2550
```

zxq82lm:
> Apologies. I hadn't noticed the visual change this makes. Is it possible for us to permit using HDR values by doing something like this:
> 
> ```diff
> -color 10.0 0.0 10.0
> +color 2550 0 2550
> ```

Technically, we could reopen RGB values by allowing something like color 2550 0 2550, but that would just keep a kind of technical debt for the sake of a third-party example.
A cleaner alternative would be to go straight to proper emissivity support in KDL. For example:
```
color orange emissive=0.3
```
This way, we introduce a dedicated field, propagate it through the engine, and achieve a luminous effect without any hacks.

shanecelis:
Let's punt on emissive for now and put that into a new issue.

shanecelis (approved):

zxq82lm:
See also: https://github.com/elodin-sys/elodin/issues/200

PR #198, opened 2025-10-06
Fix db connection thread leak

anderspitman:
When connected to an elodin-db instance, if the connection is dropped, the thread managing the connection was going into a tight loop. This commit changes it to exit when recv returns an error, which results in the loop breaking and the thread exiting.

### DIFF
--- a/libs/elodin-editor/src/ui/actions.rs
+++ b/libs/elodin-editor/src/ui/actions.rs
@@ -37,20 +37,18 @@ impl LuaActor {
                 warn!(?err, "error spawning lua client");
                 return;
             }
-            loop {
-                if let Ok((cmd, res_tx)) = cmd_rx.recv() {
-                    match lua.load(&cmd).eval_async::<MultiValue>().await {
-                        Ok(values) => {
-                            let val = values
-                                .iter()
-                                .map(|value| format!("{:#?}", value))
-                                .collect::<Vec<_>>()
-                                .join("\t");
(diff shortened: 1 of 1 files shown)


libs/elodin-editor/src/ui/actions.rs
```diff
@@ -37,20 +37,18 @@ impl LuaActor {
...
-                                .iter()
-                                .map(|value| format!("{:#?}", value))
-                                .collect::<Vec<_>>()
-                                .join("\t");
-                            let _ = res_tx.send(Ok(val));
-                        }
-                        Err(err) => {
-                            let _ = res_tx.send(Err(err.to_string()));
-                        }
+            while let Ok((cmd, res_tx)) = cmd_rx.recv() {
```
anderspitman: This should be the only line that actually changed. The others are indentation.

shanecelis: Should we consider logging a warning when this error breaks this loop?

shanecelis (approved):
Approved with one log for consideration.

libs/elodin-editor/src/ui/actions.rs
shanecelis: Maybe something like this won't break things up:

```rust
cmd_rx.recv().inspect_err(|e| warn!("command receive error {e}")
```

anderspitman: I like this idea. I'll add it in.

PR #201, opened 2025-10-07
feat: add 3D ellipsoid mesh & color transparency support

zxq82lm:


zxq82lm:
@codex review

shanecelis:
I confirm the ellipse is shown in the simulation when included in KDL.

Screenshot 2025-10-08 at 2 48 05 PM

shanecelis:
If I make my ellipse big enough like 10,000 times bigger,
```
    ellipsoid scale="10000*(2.9, 0.5, 0.5)" {
        color green 120
    }
```
it is no longer visible until I zoom out. Suggested fix: Draw the front _and_ back faces of the ellipse. That might be the easiest thing.

libs/elodin-editor/src/object_3d.rs
```diff
@@ -188,32 +202,169 @@ pub fn compile_eql_expr(expression: eql::Expr) -> CompiledExpr {
     }
 }
 
+fn parse_scale_literal(scale: &str) -> Option<CompiledExpr> {
```
shanecelis: Do we have special parsing like this for the `pos` attribute, e.g., `pos="rocket.world_pos + (0, 0, 0, 0, 5, 0 1)"`? I thought a `pos` was just EQL and that's what a scale could be too.

libs/elodin-editor/src/object_3d.rs
```diff
@@ -188,32 +202,169 @@ pub fn compile_eql_expr(expression: eql::Expr) -> CompiledExpr {
...
+        let value = expr
+            .execute(entity_map, component_value_maps)
+            .map_err(|err| err.to_string())?;
+        component_value_to_vec3(&value)
+    } else {
+        Ok(Vec3::ONE)
+    }
+}
+
+fn component_value_to_vec3(value: &ComponentValue) -> Result<Vec3, String> {
```
shanecelis: Suggest making this a proper error type, new or added to some relevant error type. It's possible this could be invoked in a tight loop, which would make these string allocations less than ideal.

libs/elodin-editor/src/object_3d.rs
```diff
@@ -188,32 +202,169 @@ pub fn compile_eql_expr(expression: eql::Expr) -> CompiledExpr {
...
+                    "scale expression must yield at least 3 values, got {}",
+                    data.len()
+                ))
+            }
+        }
+        _ => Err("scale expression must yield an F32 or F64 array".to_string()),
+    }
+}
+
+pub fn apply_ellipsoid_background(
```
shanecelis: This doesn't seem to be working. I suggested a different possible fix by drawing the backfaces of the ellipse. If that works, maybe we can drop this idea of changing the background color.

libs/elodin-editor/src/object_3d.rs
```diff
@@ -238,64 +389,127 @@ pub fn create_object_3d_entity(
...
+                AlphaMode::Blend
+            } else {
+                AlphaMode::Opaque
+            };
+
+            let material_handle = material_assets.add(StandardMaterial {
+                base_color: bevy_color,
+                alpha_mode,
+                unlit: false,
+                double_sided: true,
```
shanecelis, line 447: Strange, you do have it marked `double_sided`. Try setting `cull_mode: None` also.

libs/elodin-editor/src/object_3d.rs
zxq82lm: - Removal of _parse_scale_literal_: only proper EQL tuples are supported for scale.
- Previously (in this PR) non-EQL expressions like "a b c" were parsed and re-evaluated when the initial EQL evaluation failed. 
- With this change, scale **must** be defined directly as a tuple "(a, b, c)" in KDL, ensuring immediate EQL
compatibility.

libs/elodin-editor/src/object_3d.rs
zxq82lm: Use dedicated ScaleEvalError: no more intermediate string allocations (format!) inside the loop.

libs/elodin-editor/src/object_3d.rs
zxq82lm: ellipsoid scale="1000 * (2.9, 0.5, 0.5)", works now:
- with cull_mode: None to the ellipsoid material combined with double_sided: true

libs/elodin-editor/src/object_3d.rs
zxq82lm: Thanks a lot for the advice, @shanecelis !

`ellipsoid scale="10000 * (2.9, 0.5, 0.5)"` now works:
- with `cull_mode: None` on the ellipsoid material, combined with `double_sided: true`.

zxq82lm:
Thanks a lot for the advice, @shanecelis !

`ellipsoid scale="10000 * (2.9, 0.5, 0.5)"` now works:
- with `cull_mode: None` on the ellipsoid material, combined with `double_sided: true`.
Screenshot 2025-10-08 at 17 20 12

Issue #192, opened 2025-10-03
bug: Editor crashes when KDL is invalid.


shanecelis:
# Problem

The editor crashes when the KDL is invalid. This is especially problematic when it's run as an application for macOS or Windows because there is no log.

# Suggested Fix

Do not panic on bad KDLs. Present a dialog that shows the error message.

# Source

This issue came from Sung.

## Mea culpa

I wanted it to panic if the KDL was invalid because there was no clear path to fixing that KDL if it came from the sim and not a file. One would have to change their simulation code. If it came from a file, it logs an error, which is not sufficient because users that didn't start their editor from the CLI won't see those errors. I was wrong to panic; I apologize.

PR #202, opened 2025-10-07
Fix the invalid KDL panic

shanecelis:
This is a fix for issue #192. The editor no longer panics on bad KDL. In addition it will show the relevant error in a dialog. A further goal is even without a log (running from GUI not command prompt), the user ought to have enough information to fix their KDL problem.

### DIFF
--- a/libs/impeller2/kdl/src/de.rs
+++ b/libs/impeller2/kdl/src/de.rs
@@ -1510,11 +1510,12 @@ graph "value" {
 "#;
         let err = parse_schematic(kdl).unwrap_err();
         assert_eq!("Unknown node type 'blah'", format!("{}", err));
-        let reporter = GraphicalReportHandler::new();
+        let reporter =
+            GraphicalReportHandler::new_themed(miette::GraphicalTheme::unicode_nocolor());
         let mut b = String::new();
-        reporter.render_report(&mut b, &err);
+        reporter.render_report(&mut b, &err).unwrap();
         assert_eq!(
-            "\u{1b}[31mkdl_schematic::unknown_node\u{1b}[0m\n\n  \u{1b}[31m×\u{1b}[0m Unknown node type 'blah'\n   ╭─[2:1]\n \u{1b}[2m1\u{1b}[0m │ \n \u{1b}[2m2\u{1b}[0m │ blah\n   · \u{1b}[35;1m──┬─\u{1b}[0m\n   ·   \u{1b}[35;1m╰── \u{1b}[35;1munknown node\u{1b}[0m\u{1b}[0m\n \u{1b}[2m3\u{1b}[0m │ graph \"value\" {\n   ╰────\n",
+            "kdl_schematic::unknown_node\n\n  × Unknown node type 'blah'\n   ╭─[2:1]\n 1 │ \n 2 │ blah\n   · ──┬─\n   ·   ╰── unknown node\n 3 │ graph \"value\" {\n   ╰────\n",
             &b
         );
     }


shanecelis:
We no longer crash and we present the error in a dialog that looks like this:

Screenshot 2025-10-09 at 2 30 06 PM

And the console message has likewise been improved from its original:
```
schematic error: UnknownNode { node_type: "boo", src: "hsplit {\nboo\n    vsplit 
share=0.1398322731256485 {\n        tabs {\n            graph \"rocket.fin_control_trim[0]\"
 name=\"Trim Control\" {\n                color 106 155 165\n            }\n        }\n 
       tabs {\n            graph \"rocket.fin_deflect[0]\" name=\"Fin Deflection\" {\n 
               color 106 155 165\n            }\n        }\n        tabs {\n    
        graph \"rocket.aero_coefs[0],rocket.aero_coefs[1],rocket.aero_coefs[2],rocket.aero_
```
to this:
```
2025-10-09 14:30:04.545 ERROR Invalid schematic for "rocket.kdl" report=kdl_schematic::unknown_node

  x Unknown node type 'boo'
   ,-[2:1]
 1 | hsplit {
 2 | boo
   : ^|^
   :  `-- unknown node
 3 |     vsplit share=0.1398322731256485 {
   `----
```

shanecelis:
# How to Exercise this Change

Run the rocket example.

```sh
cd libs/nox-py
cargo run --manifest-path=../../apps/elodin/Cargo.toml editor examples/rocket.py
```

* Hit Command-P, Save Schematic.
* Edit "rocket.kdl" and add something not well formed.
* Observe error dialog (either on reload or rerun)

shanecelis:
@codex review this PR.

shanecelis:
@codex, review this PR, please.

libs/elodin-editor/src/ui/modal.rs
```diff
@@ -217,25 +331,23 @@ impl WidgetSystem for ModalUpdateGraph<'_, '_> {
...
 
-                        for (component_id, _) in components.0.iter() {
-                            let Some(metadata) = metadata_store.get_metadata(component_id) else {
-                                continue;
-                            };
+                        for (component_id, _, metadata) in &available_components {
                             ui.selectable_value(
                                 m_component_id,
-                                Some(*component_id),
+                                Some(**component_id),
```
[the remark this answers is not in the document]
shanecelis: I don't see that here.

libs/elodin-editor/src/ui/button.rs
```diff
@@ -378,7 +378,7 @@ impl EButton {
             ui.painter()
                 .layout_no_wrap(self.label.to_string(), font_id.clone(), self.color);
 
-        let desired_width = self.width.unwrap_or(ui.available_width());
+        let desired_width = self.width.unwrap_or(galley.size().x + self.margin.sum().x);
```
anderspitman, line 381: It's unclear to me what the effect of this change will be. Is it directly related to this PR?

libs/elodin-editor/src/ui/modal.rs
```diff
@@ -3,21 +3,96 @@ use bevy::{
...
     },
+    prelude::{Children, In},
     window::Window,
 };
 use bevy_egui::{EguiContexts, egui};
-use impeller2_bevy::{ComponentMetadataRegistry, ComponentPath, ComponentPathRegistry};
+use impeller2::types::ComponentId;
+use impeller2_bevy::{ComponentMetadataRegistry, ComponentPathRegistry};
+
+// Modal system for displaying dialogs and messages.
```
anderspitman, line 13: This is very helpful

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -205,11 +207,45 @@ impl Plugin for UiPlugin {
 pub enum SettingModal {
```
anderspitman, line 207: This is more out of personal curiosity but why is `SettingModal` defined here instead of in `modal.rs`?

libs/elodin-editor/src/ui/modal.rs
```diff
@@ -123,8 +226,10 @@ impl WidgetSystem for ModalUpdateGraph<'_, '_> {
             entities_meta,
             mut setting_modal_state,
             metadata_store,
-            path_reg,
+            path_reg: _path_reg,
```
anderspitman, line 229: I'm assuming these were linting warnings?

libs/elodin-editor/src/ui/modal.rs
```diff
@@ -163,7 +268,7 @@ impl WidgetSystem for ModalUpdateGraph<'_, '_> {
 
         let selected_entity = entities_meta
             .iter()
-            .find(|(entity_id, _, _, _)| m_entity_id.is_some_and(|eid| eid == **entity_id));
+            .find(|(entity_id, _, _, _)| m_component_id.is_some_and(|eid| eid == **entity_id));
```
anderspitman, line 271: I don't understand this change, I think primarily because I don't know what a component_id is in our system. I understand entity_id from an ECS perspective.

Regardless this change seems to be mixing the two.

libs/impeller2/kdl/src/de.rs
```diff
@@ -1499,4 +1499,24 @@ graph "value" {
...
+}
+"#;
+        let err = parse_schematic(kdl).unwrap_err();
+        assert_eq!("Unknown node type 'blah'", format!("{}", err));
+        let reporter =
+            GraphicalReportHandler::new_themed(miette::GraphicalTheme::unicode_nocolor());
+        let mut b = String::new();
+        reporter.render_report(&mut b, &err).unwrap();
+        assert_eq!(
+            "kdl_schematic::unknown_node\n\n  × Unknown node type 'blah'\n   ╭─[2:1]\n 1 │ \n 2 │ blah\n   · ──┬─\n   ·   ╰── unknown node\n 3 │ graph \"value\" {\n   ╰────\n",
```
anderspitman, line 1518: Is this test going to be dependent on miette implementation details? Maybe not a problem since we'd only need to fix it when updating miette.

libs/elodin-editor/src/ui/button.rs
shanecelis: It isn't directly related. I found that our `EButton` when you would try to lay it out with `horizontal`, the first button would always take up all the space. This was my attempt to make it play nicely with its neighbors and it seems to work, but it was an AI suggested fix.

libs/elodin-editor/src/ui/mod.rs
shanecelis: That was Andrei's call initially. But I'd prefer those be moved now that I look at it.

libs/impeller2/kdl/src/de.rs
shanecelis: You're right. It's a brittle test. Not much better than a golden file. However, sometimes with tests, you'll only test the fruit of the tree, and you don't capture what the tree looks like. This my attempt to capture, here's what it looked like at _this_ time. I wouldn't want many of these tests, but if this one fails when we update miette, then it shows something has changed and I'm ok with that.

Issue #205, opened 2025-10-08
Need documentation on requirement for git-lfs when building from repo


shanecelis:
We use git-lfs for a few files:

```sh
$ git-lfs ls-files
2ad6982458 * assets/edu-450-v2-drone.glb
33efc31be5 * assets/oresat-low.glb
0a9cfdf279 * assets/talon-quad-v2.glb
1e8e3c1994 * assets/talon-quad.glb
57cb887c62 * libs/elodin-editor/src/assets/diffuse.ktx2
1194177a15 * libs/elodin-editor/src/assets/fonts/IBMPlexMono-Medium_ss04.ttf
158872488c * libs/elodin-editor/src/assets/specular.ktx2
```
We can do builds, but the builds are bad because the fonts and embedded assets are actually git-lfs pointers not the files themselves. So it may behoove us to check them during compilation instead of producing bad builds.

# TODO
- [ ] Document requirement
- [ ] Compile error on git-lfs not present or git-lfs pointer files.

shanecelis:
Don't close until the git-lfs issue is fixed on CI.

shanecelis:
Issue #208 covers the git-lfs issue. This issue is for docs and it is mended.

PR #206, opened 2025-10-08
build(editor): Ensure LFS file isn't an LFS file pointer.

shanecelis:
This ensures a particular LFS file, a font file in this case, is not a LFS pointer file. This should protect us against producing bad builds when git-lfs has not been installed.

This also adds `git-lfs` to the list of required packages.

This fixes issue #205.

## Acceptance Criteria
- Builds when git-lfs is present and lfs files have been downloaded.
- Fails when git-lfs is not present and lfs files are pointers with an error about LFS.

### DIFF
--- a/libs/elodin-editor/src/ui/actions.rs
+++ b/libs/elodin-editor/src/ui/actions.rs
@@ -37,7 +37,10 @@ impl LuaActor {
                 warn!(?err, "error spawning lua client");
                 return;
             }
-            while let Ok((cmd, res_tx)) = cmd_rx.recv().inspect_err(|e| warn!("command receive error: {e}")) {
+            while let Ok((cmd, res_tx)) = cmd_rx
+                .recv()
+                .inspect_err(|e| warn!("command receive error: {e}"))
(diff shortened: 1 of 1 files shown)


shanecelis:
I updated the documentation that explains how to install git-lfs. I also included an automatic check during the build that will detect if you haven't installed git-lfs. Unfortunately, our CI build system on buildkite we've come to find DOES NOT use git-lfs either! So that check has been disabled. However, for the purposes of testing, I recommend removing that check and exercising this against a repo with and without lfs enabled to ensure it fails in the case with an error about git-lfs when git-lfs is not installed.

I noted the CI build issue in #208, and we should remove the exclusion of this test when that's resolved.

shanecelis:
@codex, review this PR.

KushMahajan (approved):
Verified the updated Git LFS installation and documentation, large file fetching works as expected. The example simulation builds and runs without issues. No errors encountered; approved to merge.
Screenshot from 2025-10-10 10-44-34

Issue #167, opened 2025-09-26
Save DB from the UI


shanecelis:
# Need

Users want to save the database from the UI or programmatically.

# Problem

Users can't save the database from the UI or programmatically.

# Solution

Let the user specify the file they want to save programmatically like so:

```py
world.run(save_db="file.db")
```

Add a command palette "Save DB As..." that will ask the user for a file name and save the current DB (or a copy) there.

# Source

Sung provided this issue.

zxq82lm:
@shanecelis I suggest adding these options directly to the command palette:
```
- Save DB (Parquet)
- Save DB as…
   - Arrow
   - CSV (lossy)
```
This way, the default Parquet save is a one-step action, while alternative formats remain accessible under the "Save DB as…" submenu.

shanecelis:
This is good to go over. I hadn't actually considered doing any "Save DB as…" conversions, but I imagine they would be a welcome addition. I wonder how much work they will be. I expect it may be a lot; we can push that into another ticket. I know they've got scripts that can do the conversions from the native DB format already.

Given this, I'd prefer it to look like this:
```
- Save DB… (MUST)
- Save DB as… (NICE TO HAVE?)
   - Parquet 
   - Arrow
   - CSV (lossy)
```
where we "Save DB…" save it in its native format. You may not have seen the DB's native format yet; it's a directory named 'db' with a lot of hash filenames (almost like git).

## Where to find the native temporary DBs we create all the time

I added a log (not present in the main branch) to show where the DBs are stored for a given run:

```
2025-10-03 16:39:29.000  INFO elodin::world_builder: in thread create db at "/var/folders/wv/4qnq4g1j3kg5gj5q6jzq4s7h0000gn/T/.tmpTcCND6/db"
```

So if you do a run, you can find the database files that were used by doing the following:
```sh
$ find /var/folders -name db -printf "%T+ %p\n" 2>/dev/null | sort | tail -2
2025-10-03+14:23:59.2291528710 /var/folders/wv/4qnq4g1j3kg5gj5q6jzq4s7h0000gn/T/.tmpfwX5Fr/db
2025-10-03+16:39:29.0083989630 /var/folders/wv/4qnq4g1j3kg5gj5q6jzq4s7h0000gn/T/.tmpTcCND6/db
```

PR #207, opened 2025-10-08
Save elodin-db from UI [abandoned]

zxq82lm:
## Description

Description in this issue https://github.com/elodin-sys/elodin/issues/167, which outlines the feature both in the Command Palette (UI) and from the Elodin CLI (programmatically). _This PR covers only “Save from UI”._

---

### Allowed Save Locations

- Relative subdirectory of the current working directory (workspace) that does not yet exist, for example: `run-2024`
- Nested path under the current working directory, as long as it stays within that tree, e.g., `exports/dbs/session_01`
-  Absolute paths (`/tmp/foo`, `C:\foo`) and upward-traversing paths (`../foo`) are not accepted.
- The final folder name must not be `db`.

### How-to test

- Run the drone example: `cargo run --manifest-path=apps/elodin/Cargo.toml editor examples/drone/main.py`
- CMD+P
- `Save DB…`
- Enter "current_run" as directory name.
- Check the content of this new directory.

### DIFF
--- a/libs/db/src/lib.rs
+++ b/libs/db/src/lib.rs
@@ -134,6 +134,61 @@ impl DB {
         db_state.write(self.path.join("db_state"))
     }
 
+    pub fn save_native_to(&self, target_db_path: PathBuf) -> Result<PathBuf, Error> {
+        let (parent_dir, final_db_dir) = if target_db_path
+            .file_name()
+            .and_then(|n| n.to_str())
+            .map(|n| n.eq_ignore_ascii_case("db"))
+            .unwrap_or(false)
+        {
+            (
+                target_db_path
+                    .parent()
+                    .map(Path::to_path_buf)
+                    .unwrap_or_else(|| PathBuf::from(".")),
+                target_db_path.clone(),
+            )
+        } else {
+            (target_db_path.clone(), target_db_path.join("db"))
+        };
+
+        let run_dir = parent_dir.clone();
+
+        if self.path.starts_with(&final_db_dir) || final_db_dir.starts_with(&self.path) {
+            return Err(Error::Io(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                "target directory overlaps database path",
+            )));
+        }
+
+        if run_dir
+            .file_name()
+            .and_then(|n| n.to_str())
+            .map(|n| n.eq_ignore_ascii_case("db"))
+            .unwrap_or(false)
+        {
+            return Err(Error::Io(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                "directory name \"db\" is not allowed",
+            )));
+        }
+
+        if run_dir.exists() || final_db_dir.exists() {
+            return Err(Error::Io(std::io::Error::new(
+                std::io::ErrorKind::AlreadyExists,
+                "target directory already exists",
+            )));
+        }
+
+        std::fs::create_dir_all(&run_dir)?;
+        let tmp_db_dir = run_dir.join("db.tmp");
+        let _ = std::fs::remove_dir_all(&tmp_db_dir);
+        std::fs::create_dir_all(&tmp_db_dir)?;
+        copy_dir_recursively(&self.path, &tmp_db_dir)?;
+        std::fs::rename(&tmp_db_dir, &final_db_dir)?;
+        Ok(final_db_dir)
+    }
+
     pub fn open(path: PathBuf) -> Result<Self, Error> {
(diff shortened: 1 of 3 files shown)


shanecelis:
Ok, I don't mind the idea of doing this as CLI argument. I'm curious though if I provide a CLI argument, is the DB created there? Or is a temporary one still created and I'm making a copy of it at some point?

For an earlier request, I had tracked down where the temp files are defined just to add some logging. Here's the snippet:
```diff
diff --git a/libs/nox-py/src/world_builder.rs b/libs/nox-py/src/world_builder.rs
index 17d52c72..4bd25ecc 100644
--- a/libs/nox-py/src/world_builder.rs
+++ b/libs/nox-py/src/world_builder.rs
@@ -258,8 +258,11 @@ impl WorldBuilder {
                 py.allow_threads(|| {
                     stellarator::run(|| {
                         let tmpfile = tempfile::tempdir().unwrap().into_path();
+
+                        let db_file = tmpfile.join("db");
+                        tracing::info!("in thread create db at {:?}", db_file.display());
                         nox_ecs::impeller2_server::Server::new(
-                            elodin_db::Server::new(tmpfile.join("db"), addr).unwrap(),
+                            elodin_db::Server::new(db_file, addr).unwrap(),
                             exec,
                         )
                         .run_with_cancellation(|| {
@@ -348,7 +351,9 @@ impl WorldBuilder {
         let mut exec = exec.compile(client.clone())?;
         let db_dir = tempfile::tempdir()?;
         let db_dir = db_dir.into_path();
-        let db = elodin_db::DB::create(db_dir.join("db"))?;
+        let db_file = db_dir.join("db");
+        tracing::info!("create db at {:?}", db_file.display());
+        let db = elodin_db::DB::create(db_file)?;
         nox_ecs::impeller2_server::init_db(&db, &mut exec.world, Timestamp::now())?;
         Ok(Exec { exec, db })
     }
```

zxq82lm:
@shanecelis: 
We therefore have two implementation options:

---

- Option A: define the database location. This is not really a “save,” but rather the definition of where the database will live and persist after the run.
```bash
cargo run --manifest-path apps/elodin/Cargo.toml -- run libs/nox-py/examples/rocket.py --set-db-directory demo-run
```

---

- Option B: perform a true save in the target directory (equivalent to “Save from UI” in this branch, where part of the code will likely be reused).
```bash
cargo run --manifest-path apps/elodin/Cargo.toml -- run libs/nox-py/examples/rocket.py --save-db demo-run
```
The timing of the snapshot is important here: it could be at the end of the simulation, for example when `max_ticks=xxx` defined in `world.run(...)` is reached, or at the end imposed by the user (CTRL+C). I’m not entirely sure we can easily have a clean save callback in that case, but I can investigate it.

---

- I hadn’t considered option A: it’s “cleaner” than option B (and probably simpler to implement). The question is whether it matches the expectations of the FSW team?
- I can implement both options A and B, or just one of them, whichever you prefer?

zxq82lm:
@x46085 thanks :)
- I replaced the set-db-directory flag with set-db.
- Regarding validations:
   - For Save from UI (Command palette), I prevent saving into a directory that already exists, and also disallow choosing the name "db" to avoid ending up with "db/db" (though I’m not sure that check is very relevant).
   - For the Elodin CLI feature, for now we don’t enforce any checks.

Do you have any suggestions for validations to add? Should they be the same as the ones used for Save from UI?

x46085:
Yea I dunno, I feel like the CLI has some footguns here, especially since people have also asked for the ability to run a simulation from an existing database, and in that case start the simulation from that point as an appending simulation run. 

Knowing nothing about this command, my expectations is it would maintain the same database name throughout, but it’s not clear if it’ll be a fresh copy from zero each time, or a continuation. 

Also more confusion will come from the current behavior of elodin-db, where you can start the db with a cli command, and then connect to it with the editor, and it’s similar appearance to this command but different behavior could be confusing.

I don’t claim to have an answer here, I defer to Shane. I think conflating the UI and CLI workflows might just be opening a can of worms, that you could avoid by just shipping the UI case first and circling back.

zxq82lm:
Ok, thanks for the perspective. We’ll discuss it with Shane tomorrow. It’s indeed a possibility to only keep “Save from UI” in this PR and continue thinking about the Elodin CLI flags independently from the UI features.

zxq82lm:
@codex analyze this pr

zxq82lm:
See https://github.com/elodin-sys/elodin/pull/233

PR #210, opened 2025-10-09
Restore vector arrow schematic support

joelreymont:
## Summary
- keep vector arrow schematic support by importing impeller2 bevy/wkt types instead of the removed impeller crate
- adjust gizmo configuration to the current bevy gizmo API and pull in the WorldPos extension trait
- include the nox ArrayBuf trait so vector arrow helpers compile

## Testing
- cargo check -p elodin-editor *(fails: missing `alsa` system library in container)*

------
https://chatgpt.com/codex/tasks/task_b_68e763d855f083278c7f29b531be44e5

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,21 +25,38 @@ This monorepo contains the source code for all Elodin simulation and flight soft
 ## Dependencies
 
 Rust 1.85.0
-Preference for Arm based Macs
+Preference for Arm-based Macs
 
-### macOS
+## Building
 
-Validated on M1 architecture, macOS 15.1.1 on 2025-08-26.
+> [!NOTE]
+> These instructions were validated on M1 architecture, macOS 15.1.1 on 2025-08-26.
+
+### Before Cloning
+
+Before cloning the source, ensure `git-lfs` is installed. 
+
+``` sh
+brew install gstreamer python gfortran openblas uv git-lfs;
+git lfs install; # This is idempotent; you can run it again.
+```
+
+### Source
+
+``` sh
+git clone https://github.com/elodin-sys/elodin.git
+```
+
 
 ### Build & Run
 ```sh
+cd elodin
 just install
 ```
 
 #### Elodin App & SDK Development
 (See [apps/elodin/README.md](apps/elodin/README.md))
 ``` sh
-brew install gstreamer python gfortran openblas uv
 
 cd libs/nox-py
 uv venv --python 3.12
--- a/docs/public/content/reference/migration/to-0.15.md
+++ b/docs/public/content/reference/migration/to-0.15.md
@@ -74,7 +74,7 @@ All the preceding information can now be specified via schematics.
         }
         object_3d ball.world_pos {
             sphere radius=0.2 {
-                color 25 50 255
+                color 25 50 255 128
             }
         }
         line_3d ball.world_pos line_width=2.0 {
@@ -87,7 +87,7 @@ The old code is python, and the new code is specified in a [KDL](https://docs.rs
 
 
 ### `el.Material.color` to `color`
-A color can be specified by its red, green, blue, and optionally its alpha components as an integer from 0 to 255.
+A color can be specified by its red, green, blue, and optionally its alpha components as integers from 0 (fully transparent) to 255 (fully opaque).
 
 OLD
 ```python
@@ -95,8 +95,18 @@ el.Material.color(0.1, 0.2, 1.0))
 ```
 NEW
 ```python
-color 25 50 255
+color 25 50 255 128
 ```
+
+Omit the fourth value if you want the color to remain fully opaque. The same syntax works with named colors:
+
+```
+    graph "value" {
+        color yolk 120
+    }
+```
+
+Here `yolk` is the base color and `120` sets a partially transparent alpha.
 
 It can also be specified by a few names:
 
@@ -122,6 +132,19 @@ It is often specified as a child of what its describing, e.g., this sphere is re
         color reddish
     }
 ```
+
+### `object_3d` ellipsoid
+An `ellipsoid` block renders an ellipsoidal volume aligned with an entity's `world_pos`. Provide a scale expression that evaluates to three components (literal values or an EQL expression) and an optional translucent color.
+
+```
+    object_3d drone.world_pos {
+        ellipsoid scale="(0.75, 0.35, 0.95)" {
+            color 255 255 0 120
+        }
+    }
+```
+
+Because the scale expression is re-evaluated every frame, you can reference live data such as `rocket.ellipsoid_scale` to animate the ellipsoid.
 
 ### `el.Panel.hsplit` to `hsplit`
 OLD
--- a/libs/elodin-editor/build.rs
+++ b/libs/elodin-editor/build.rs
@@ -1,0 +1,72 @@
+use std::fs::File;
+use std::io::Read;
+use std::path::Path;
+
+fn main() {
+    // We're picking one file out of the handful of git LFS files present. It is
+    // the smallest.
+    let path = Path::new("src/assets/fonts/IBMPlexMono-Medium_ss04.ttf");
+
+    // You can create an LFS pointer file to test this code. You can't check it
+    // in though without upsetting git-lfs.
+    //
+    // ```sh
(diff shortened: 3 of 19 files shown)


libs/elodin-editor/src/plugins/gizmos.rs
```diff
@@ -98,6 +105,35 @@ fn render_vector_arrow(
...
+            if let Ok(origin_value) = origin_expr.execute(&entity_map, &values) {
+                if let Some(origin) = component_value_tail_to_vec3(&origin_value) {
+                    start = origin;
+                }
+            }
+        }
+
+        let end = start + direction;
+        let color = bevy::prelude::Color::srgb(arrow.color.r, arrow.color.g, arrow.color.b);
+        gizmos.arrow(start, end, color);
```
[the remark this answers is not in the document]
joelreymont: @codex Fix this!

joelreymont:
This is based off of PR #209

joelreymont:
@codex review this PR

shanecelis:
The arrows are cool!

Screenshot 2025-10-09 at 3 29 07 PM

I found one issue. I'd like for the scale to be able to accept integers or floats. Right now if you provide 'scale=2', then it acts like you've provided nothing. Also it'd be nice to see an error or warning log when an invalid type for `scale` is given.

shanecelis:
Can we separate out all the version bumps in the Cargo files into a separate PR? I understand it's probably required for your local environment, but it's not germane to this feature, is it?

joelreymont:
> Can we separate out all the version bumps in the Cargo files into a separate PR? I understand it's probably required for your local environment, but it's not germane to this feature, is it?

I can't and not by lack of trying! I'm on macOS Tahoe 26.01 on all my Macs and get this error even if I build in a Nix dev shell

```
2025-10-10 09:00:14.441  INFO ELODIN_ASSETS_DIR defaulted to "assets"

thread 'main' panicked at /Users/joel/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/objc2-foundation-0.2.2/src/generated/NSEnumerator.rs:7:1:
invalid message send to -[_TtGCs23_ContiguousArrayStorageCSo8NSScreen_$ countByEnumeratingWithState:objects:count:]: expected return to have type code 'q', but found 'Q'
```

If you can build and run with the version bumps then I suggest we keep them, otherwise users with a fresh Homebrew or those on macOS Tahoe won't be able to run the editor.

joelreymont:
> Also it'd be nice to see an error or warning log when an invalid type for `scale` is given.

Your wish is [my command](https://github.com/elodin-sys/elodin/pull/210/commits/b6c5ed9750b6509c714dce639c8bccea2cac792c)!

shanecelis:
I am seeing an odd visual artifact, the blue line in the center of the screen. Made odder still by the fact I have determined that there is one `gizmo.arrow()` call per frame, and when that call is removed this artifact goes away. Honestly, it looks like a Bevy bug to me right now, but I didn't find it in their issues.

https://github.com/user-attachments/assets/12291720-c547-4449-8cd5-4a2bd78bcd5d

If I were to guess, I'd say this was an arrow head line that was supposed to be attached to the vector and got misplaced.

I am running the following:
```sh
cd libs/nox-py
cargo run --features elodin-editor/inspector --manifest-path=../../apps/elodin/Cargo.toml editor ../../examples/ball/main.py
```

shanecelis:
It's locked to the center of the screen. Not even the viewport.

https://github.com/user-attachments/assets/4aea41aa-7e26-4ded-bce2-56ac2e50591d

shanecelis:
Benoît has the gizmo from another camera issue sorted out. He's going to push directly into this branch and shepherd this over the finish line.

I noticed one issue that may or may not be related to this PR, but when I do a Save Schematic from the ball example. I get this KDL, which looks okay:

```
tabs {
    viewport name=Viewport pos="(0,0,0,0, 8,2,4)" look_at="(0,0,0,0, 0,0,3)"
}
object_3d ball.world_pos {
        color 255 128 0
    }
}
    color 255 255 255
}
vector_arrow "ball.world_vel[3],ball.world_vel[4],ball.world_vel[5]" origin=ball.world_pos name="Ball Velocity" {
    color 0 0 255
}
```
However, when it is loaded, the viewport no longer shows the grid. In the sim code it has 'grid=#true'. Here the grid is missing from the saved KDL.

zxq82lm:
Commit: https://github.com/elodin-sys/elodin/pull/210/commits/d2499fb095cb928fee18333ab1f53b21a3315b29
- `libs/elodin-editor/src/plugins/gizmos.rs`: added guards that drop near-zero arrows and converted world origins to grid-space via FloatingOriginSettings, which stops zero-velocity samples from emitting the blue artifact. The gizmo config now pins its render_layers to a dedicated constant (GIZMO_RENDER_LAYER) so we can segregate gizmo drawing from UI cameras.
- `libs/elodin-editor/src/plugins/navigation_gizmo.rs`: updated the render-layer allocator so the gizmo layer is pre-reserved; this avoids collisions when navigation gizmos request layers from the pool.
- `libs/elodin-editor/src/ui/tiles.rs`: ensured every spawned 3D viewport camera opts into the gizmo layer, keeping gizmos visible in the editor while the UI Camera2d remains isolated on its own layer.
- `libs/elodin-editor/src/ui/plot_3d/mod.rs`: switched Line3d trails to render exclusively on the gizmo layer, preventing the default UI camera from redrawing the trajectory a second time.

image

zxq82lm:
> I noticed one issue that may or may not be related to this PR, but when I do a Save Schematic from the ball example. I get this KDL, which looks okay:
> 
> ```
> tabs {
>     viewport name=Viewport pos="(0,0,0,0, 8,2,4)" look_at="(0,0,0,0, 0,0,3)"
> }
> object_3d ball.world_pos {
>     sphere radius=0.20000000298023224 {
>         color 255 128 0
>     }
> }
> line_3d ball.world_pos line_width=2.0 {
>     color 255 255 255
> }
> vector_arrow "ball.world_vel[3],ball.world_vel[4],ball.world_vel[5]" origin=ball.world_pos name="Ball Velocity" {
>     color 0 0 255
> }
> ```
> 
> However, when it is loaded, the viewport no longer shows the grid. In the sim code it has 'grid=#true'. Here the grid is missing from the saved KDL.

--- 

It’s an issue on main, unrelated to this PR: see https://github.com/elodin-sys/elodin/issues/211

libs/elodin-editor/src/plugins/gizmos.rs
```diff
@@ -26,15 +41,19 @@ impl Plugin for GizmoPlugin {
...
     config.enabled = true;
+    config.render_layers = RenderLayers::layer(GIZMO_RENDER_LAYER);
 }
 
 fn render_vector_arrow(
     entity_map: Res<EntityMap>,
     query: Query<(Option<&Transform>, &ComponentValueMap)>,
     arrows: Query<&VectorArrow>,
+    vector_arrows: Query<(&VectorArrow3d, &VectorArrowState)>,
+    component_values: Query<&'static WktComponentValue>,
```
shanecelis, line 55: It's become clear to me that there are two vector arrow implementations. The old one is `VectorArrow` by Sascha. The new one is `VectorArrow3d` by Joel. The new one works and does what the PR asks, so I'm inclined to use that one.

The old does not currently have a means for being exhibited. It may work once fed, and it has some distinct features that we did not ask for in the issue like, whether its "attached" or in the "body_frame". Before KDL these notions of attached and body_frame were common. After KDL they seem to have been phased out. So maybe it's for the best to let them go. Whether something is "attached" or not I believe we can express using the new `vector_arrow`. Something being in the body_frame would probably require a matrix or quaternion multiplication operation in order for us to express that in the SDL using EQL.

x46085: In game engine editors, they are always in the body frame of the game object component they are associated with, and the components would have a hierarchy. Meaning if it was associated with the root “rocket” component, it would show body frame for that, but if say you had articulated fins as children components that included a widget, the transforms automatically would be applied so they would be in body frame to the fins. That is one way to avoid having to specify this other than with the implicit component hierarchy. But for our purposes to meet Nat’s needs, enough to just keep it simple

PR #212, opened 2025-10-10
Keep viewport KDL flags consistent across schematic save/load

zxq82lm:
## Description
- Ensure viewport KDL flags are preserved when exporting schematics, and reapplied on reload, so settings like `show_grid` & `hdr` survive a “Save Schematic” roundtrip. 
- Ensure the KDL serializer always emits viewport properties in this order: `name → fov → pos → look_at → hdr → show_grid → active` (see also this test: `cargo test -p impeller2-kdl test_viewport_property_order`).

---

**How to Test**:
- Launch the editor with the ball example.
- Change in the loaded KDL in `show_grid=#true`, `show_grid=#false`, remove `show_grid`, check it is consistant (grid displayed or not), idem for `hdr=#true`, `hdr=#false`, remove `hdr`.
- Ensure the grid is visible, then save the schematic.
   - Ensure the serialized viewport KDL keeps the legacy order (`name → fov → pos → look_at → hdr → show_grid → active`) and contains `show_grid=#true` at the right place.
- Do the same kind of test for `hdr`.

### DIFF
--- a/libs/elodin-editor/src/ui/schematic/load.rs
+++ b/libs/elodin-editor/src/ui/schematic/load.rs
@@ -1,5 +1,6 @@
 use bevy::{ecs::system::SystemParam, prelude::*};
 use bevy_egui::egui::Color32;
+use bevy_infinite_grid::InfiniteGrid;
 use egui_tiles::{Container, Tile, TileId};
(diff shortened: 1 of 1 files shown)


zxq82lm:
@codex analyze this pr

Issue #217, opened 2025-10-11
There is a vestigial vector arrow implementation in the editor


shanecelis:
Re: PR #210. It's become clear to me that there are two vector arrow implementations. The old one is VectorArrow by Sascha. The new one is VectorArrow3d by Joel. The new one works and does what the PR asks, so I'm inclined to use that one.

The old does not currently have a means for being exhibited. It may work once fed, and it has some distinct features that we did not ask for in the issue like, whether its "attached" or in the "body_frame". Before KDL these notions of attached and body_frame were common. After KDL they seem to have been phased out. So maybe it's for the best to let them go. Whether something is "attached" or not I believe we can express using the new vector_arrow. Something being in the body_frame would probably require a matrix or quaternion multiplication operation in order for us to express that in the SDL using EQL.

shanecelis:
# Feature Request

Add `body_frame=#true` and `normalize=#true` support.

Starting from the ball example.

# Case 1 (supported)

```kdl
vector_arrow "ball.world_vel[3],ball.world_vel[4],ball.world_vel[5]" origin="ball.world_pos" scale=1.0 name="Ball Velocity"
```

Given ball's velocity (A) and ball's world position (B) and scale (s), this describes a vector drawn from B to B + s A.

# Case 2 (supported)

```kdl
vector_arrow "ball.world_vel[3],ball.world_vel[4],ball.world_vel[5]" scale=2
```

Given ball's velocity (A) and scale (s), this describes a vector drawn from the origin, or zero vector, to s A.

# Case 3 (requested)

```kdl
vector_arrow "ball.world_vel[4],ball.world_vel[5],ball.world_vel[6]" origin="ball.world_pos" scale=1.0 name="Ball Velocity" body_frame=#true
```

Given ball's velocity (A) and ball's world position (B_pos) and world orientation (B_quat) and scale (s), this describes a vector drawn from B_pos to B_pos + s B_quat * A * B_quat*.

# Case 4 (requested)

```kdl
vector_arrow "ball.world_vel[3],ball.world_vel[4],ball.world_vel[5]" origin="ball.world_pos" scale=1.0 name="Ball Velocity" body_frame=#true normalize=#true
```

Given ball's velocity (A) and ball's world position (B_pos) and world orientation (B_quat) and scale (s), this describes a vector drawn from B_pos to B_pos + s B_quat * normalize(A) * B_quat*.

anderspitman:
[...]

PR #223, opened 2025-10-14
Remove old arrow implementation and add `normalize` and `body_frame` (#217)

joelreymont:


joelreymont:
@codex Review this PR

joelreymont:
> ### 💡 Codex Review
> https://github.com/elodin-sys/elodin/blob/ce8722bdf53a8d2eca60ca7af2ce9ebaaf9f0269/libs/elodin-editor/src/plugins/gizmos.rs#L64-L87
> 
> 
> **P1 Badge Respect body-frame flag when rendering vector arrows**
> The renderer always treats `VectorArrow3d` vectors as world-space directions: the new loop scales the vector and immediately draws `gizmos.arrow(start, end, …)` without ever consulting `arrow.in_body_frame`. WKT parsing still populates this flag, so any arrow authored with `in_body_frame = true` (e.g. velocity vectors relative to a spacecraft) will no longer be rotated by the entity’s orientation and will point in the wrong direction as the body turns. Consider rotating the direction by the associated transform when `in_body_frame` is set or remove the flag from the schema to avoid mis-rendering.
> 
> ℹ️ About Codex in GitHub

@codex Fix this

shanecelis:
I updated the issue #217 with a few requests that will bring us up to feature parity with the prior implementation.

joelreymont:
CleanShot 2025-10-15 at 19 43 45

anderspitman:
@zxq82lm @shanecelis this is ready for review. Everything appears to be working to me. I was under the impression that normalization and rotation weren't quite working yet, so please be sure to verify this.

The only changes I've made is merging in main and adding some vector arrows to the drone example KDL.

You should be able to fully exercise the functionality by running the ball example and drone example.

examples/drone/sim.py
```diff
@@ -165,6 +165,9 @@ def world() -> tuple[el.World, el.EntityId]:
                 graph "drone.magnetometer"
             }
         }
+        vector_arrow "(1, 0, 0)" origin="drone.world_pos" scale=1.0 name="Drone Velocity X" body_frame=#true
+        vector_arrow "(0, 1, 0)" origin="drone.world_pos" scale=1.0 name="Drone Velocity Y" body_frame=#true
+        vector_arrow "(0, 0, 1)" origin="drone.world_pos" scale=1.0 name="Drone Velocity Z" body_frame=#true
```
zxq82lm, line 170: These three vector_arrow objects are not pointing to the drone’s instantaneous velocity: (1, 0, 0), (0, 1, 0), (0, 0, 1) just encode constant vectors aligned with the local X/Y/Z axes. They’re visual axes, not the velocity. 
Should we keep it?

---

If we need an arrow example involving the drone velocity, we could use this instead:
```
vector_arrow "drone.world_vel[3],drone.world_vel[4],drone.world_vel[5]" origin="drone.world_pos" scale=1.0 name="Drone Velocity" normalize=#true body_frame=#true {
            color 0 0 255
}
```
image

zxq82lm (approved):
**Test with ball example** (arrow, normalize, body_frame):
- 1/ Set up a constant slanted wind:
```python
@el.dataclass
class WindData(el.Archetype):
    seed: el.Seed = field(default_factory=lambda: jnp.int64(0))
    wind: Wind = field(default_factory=lambda: jnp.array([-5.0, 0.5, -1.2]))

def apply_drag(w: Wind, v: el.WorldVel, f: el.Force) -> el.Force:
    fluid_movement_vector = w
    fluid_movement_vector -= v.linear()
```
- 2/ Added the wind arrow (green, normalized, body-frame off):
```
vector_arrow "ball.wind[0],ball.wind[1],ball.wind[2]" origin="ball.world_pos" scale=1.0 name="Wind"   normalize=#true {
            color 0 255 0
}
```
- 3/ Validation flow: run once with `body_frame=#true` on the blue velocity arrow, then flip to `body_frame=#false`
```
vector_arrow "ball.world_vel[3],ball.world_vel[4],ball.world_vel[5]" origin="ball.world_pos" scale=1.0 name="Ball Velocity" body_frame=#true  normalize=#true {
            color 0 0 255
}
```

shanecelis:
Does anyone know what explains this lag when I scrub across the timeline?

https://github.com/user-attachments/assets/7f5b5517-9110-4236-8553-b5715b7759d1

x46085:
> Does anyone know what explains this lag when I scrub across the timeline?

Lag is typically from the shader draw call layers and them being out of sync, but this doesn't look like that. This doesn't look so much like lag as the position being multiplied by the magnitude of the direction of the object's rotation, or some similar cascading effect.

shanecelis:
This isn't a draw call or network issue in that sense of lag. I want to say it's something to do with the expression for origin is evaluated with a different or "laggy" time index, but the positions that the vector takes aren't part of its position history; they're wildly off. I wonder if this is a complication of [big_space](https://docs.rs/big_space/latest/big_space/).

shanecelis:
I don't want to hold this PR up. But I think we should file an issue against this scrub distortion. I think when that's fixed it'll also fix the jitter that vector arrows exhibit during regular play. @anderspitman, can you file the issue? You're welcome to use the above video.

@x46085, did the prior vector arrows implementation exhibit this or a jitter issue?

x46085:
> @x46085, did the prior vector arrows implementation exhibit this or a jitter issue?

No, and I’d prefer we don’t give up on it too easily. Fine to merge so long as Anders picks up working on the fix for the deeper issue as the next thing.

shanecelis:
Shall we give Anders Thursday to try to fix it?

anderspitman:
Whoa that's pretty bad. I observed this in my testing but assumed it was just a performance issue due to WSL (the GUI stuff is really slow, I'm looking into if there's a way to fix that). It seems weird to me that if you stop scrubbing the arrows "catch up" to the object. If it's due to a calculation bug or out of sync values between the arrows and the object, I would expect them to stay offset for a given timestamp once you stop scrubbing.

examples/drone/sim.py
anderspitman: Thanks for the comprehensive testing @zxq82lm! I used the axes arrows since that's what Joel had on the ball previously, but I like the velocity arrow as well. I'll defer to @x46085 on this.

anderspitman:
@x46085 @shanecelis I'm not going to have this fixed today. I'm fine merging or waiting. Either way I'm planning to pick up the jitterbug again on Friday.

shanecelis:
I think I got this sorted. I had us use doubles throughout the vector arrow computation until the end. Didn't fix it, but I'm keeping that because it's sound. I used a schematic that had the camera not moving, and there the jitter went away. What it came down to was an ordering problem in the schedules I think. Most things were positioned in `PreUpdate` and the arrows were drawn in `PostUpdate`. Anyway, it's fixed now. Yay! See for yourself:

https://github.com/user-attachments/assets/53e5cf11-428b-4aa7-bef8-821419162ed1

anderspitman:
Great work @shanecelis! I reviewed the additional changes you made and I think this is good to merge.

PR #226, opened 2025-10-14
Add plane mesh support for object_3d

joelreymont:
## Summary
- Add a plane mesh option to object_3d schematics, including parser/serializer coverage
- Render plane meshes through the editor with double-sided materials and creation/inspection UI
- Document plane usage in the migration guide

### DIFF
--- a/examples/ball/sim.py
+++ b/examples/ball/sim.py
@@ -39,7 +39,7 @@ def world(seed: int = 0) -> el.World:
         }
         object_3d "(0,0,0,1, 0,0,0)" {
             plane width=2000 depth=2000 {
-                color 32 128 32
+                color 32 128 32 125
             }
         }
     """, "ball.kdl")


joelreymont:
@codex Review this PR

joelreymont:
CleanShot 2025-10-14 at 18 38 37

x46085:
One of the use-cases for this is to create a "ground plane" that stretches to the horizon. Can you test the behavior for what this looks like when scaled massively up?


Other features worth considering:
- introducing a noise profile to create a terrain like surface
- introducing a texture or material to make it look like terrain

Fine to put these in a new ticket. Creating a realistic terrain ground plane would be a great next step / new related feature.

shanecelis:
This looks good. 

Screenshot 2025-10-14 at 10 37 21 AM

I might suggest we add a plane to the ball example.

```diff
diff --git a/examples/ball/sim.py b/examples/ball/sim.py
index bf721e3c..c204345d 100644
--- a/examples/ball/sim.py
+++ b/examples/ball/sim.py
@@ -37,7 +37,12 @@ def world(seed: int = 0) -> el.World:
         vector_arrow "ball.world_vel[3],ball.world_vel[4],ball.world_vel[5]" origin="ball.world_pos" scale=1.0 name="Ball Velocity" {
             color 0 0 255
         }
-    """)
+        object_3d "(0,0,0,1, 0,0,0)" {
+            plane width=2000 depth=2000 {
+                color 32 128 32
+            }
+        }
+    """, "ball.kdl")
     return world
 
```

libs/elodin-editor/src/ui/command_palette/palette_items.rs
```diff
@@ -1064,6 +1064,47 @@ pub fn create_3d_object() -> PaletteItem {
...
+                                            move |In(dimensions): In<String>| {
+                                                let parts: Vec<f32> = dimensions
+                                                    .split_whitespace()
+                                                    .filter_map(|s| s.parse().ok())
+                                                    .collect();
+
+                                                let (width, depth) = match parts.as_slice() {
+                                                    [w, d] => (*w, *d),
+                                                    [w] => (*w, *w),
+                                                    _ => (10.0, 10.0),
```
shanecelis, line 1091: I wonder if we should make the default bigger. Maybe 100, 100? I'd like infinite, but that doesn't look easy.

shanecelis: Actually. I retract this change. Let's leave it 10,10. 100,100 might obscure more things in weird scenarios.

PR #233, opened 2025-10-14
Add native DB snapshot support across Elodin CLI, Python & UI

zxq82lm:
> See the conversation on this abandoned PR as well: [#207](https://github.com/elodin-sys/elodin/pull/207)

## Content
Added an end-to-end “native” snapshot workflow:
- CLI/Python: expose `ArchiveFormat::Native` so `elodin-db lua` and `nox_py.Exec.save_archive(..., "native")` trigger the new snapshot mode.
- Core DB: introduce a snapshot barrier for writers, implement `DB::flush_all`, add reflink-or-sparse copy machinery, and surface a helpful error when a target directory lacks `db_state`. `save_archive` now routes Native requests through this path.
- UI: command palette gains "Save DB" with workspace-aware validation (relative paths confined to the workspace, or absolute paths forwarded verbatim) and inline error reporting.

---

## Implementation
- Extended ArchiveFormat with Native and taught the CLI/Python bindings about it (`libs/impeller2/wkt/src/msgs.rs`, `libs/nox-py/src/exec.rs`).
- Expanded the snapshot barrier to cover all DB mutations (including metadata updates) and exposed `DB::begin_snapshot` (`libs/db/src/lib.rs`).
- Implemented `DB::flush_all`, plus flush/fsync helpers on storage types (`libs/db/src/lib.rs`, `libs/db/src/time_series.rs`, `libs/db/src/msg_log.rs`, `libs/db/src/append_log.rs`).
- Built native snapshot-copy plumbing (reflink fallback to sparse copy) and routed `save_archive` to it for the Native format (`libs/db/src/lib.rs`, `libs/db/src/arrow/mod.rs`).
- Added `MissingDbState` to improve the error message when opening a directory that isn’t a database (`libs/db/src/lib.rs`, `libs/db/src/error.rs`).
- Created an async integration test ensuring the snapshot blocks inflight writes and produces a consistent clone (`libs/db/tests/src/lib.rs`).
- Wired the command palette’s “Save DB” entry to the native snapshot path with workspace-scoped validation rules, and removed the restriction that forbade paths ending in `db` (`libs/elodin-editor/src/ui/command_palette/palette_items.rs`).

### Backend Path Safeguards
- Refuses destinations that sit inside the live database directory to avoid self-nesting.
- Requires the destination directory to be absent before starting the copy.
- On non-Windows hosts, rejects Windows-style paths (`C:\...`, `\\server\share\...`) with a clear error.
- Surfaces underlying filesystem errors verbatim (read-only, missing directory, permission denied, etc.).

---

## How to test

---

### elodin-db-tests ✔️

**Test description**:
- Preparation. Create an ephemeral database, register an `archive_native_test` component, and send three values (10.5, 20.5, 30.5). Also add a 4 MiB “filler” file to slow down the native copy.
- Snapshot vs. late write. Start `SaveArchive { format: Native }` while another task attempts to inject a fourth value (99.5). The call to `zip(save_future, write_future)` forces both futures to complete and guarantees that the snapshot is taken in parallel with the write.
- Assertions on the snapshot side:
  - The response returns the requested root directory.
  - The root folder exists and opens with `DB::open`, proving that the native copy is valid.
  - The snapshot contains only the first three measurements; the value 99.5 does not appear.
- Assertions on the live database side. After the save, the original DB does contain the 99.5 sample. This proves that the barrier correctly pauses writes during the copy without losing them: the snapshot is taken on a frozen state, then the pending writes resume.

**Explanation about Futures**:
> - `save_future` sends a `SaveArchive` request to the DB server to start the native snapshot.
> - `write_future` sends a late message (a `LenPacket` with the value 99.5) to the same DB.
> - Each one triggers a network operation to the database. By starting them together (and awaiting both futures to complete), the DB receives a snapshot request and a write almost simultaneously. That’s exactly what we want to test: the barrier must prevent the write from being included in the snapshot, then let it through once the copy is finished.

**Test**:
- Run: `cargo test -p elodin-db-tests test_save_archive_native_blocks_writes`

---

#### UI Path Validation Rules
- Accepts either a workspace-relative path or an absolute path supplied by the user.
- Ensures relative paths stay within the workspace root; any escape (`../`) is rejected immediately.
- Rejects Windows-style prefixes (`C:\`, `\\server\share`) when connected to a non-Windows database host.
- Blocks submission if the target folder already exists (to avoid overwriting snapshots).
- Displays backend error responses verbatim when the server rejects the path (permission issues, missing dirs, etc.).

> Note: The backend remains the final arbiter, but this UI-side filtering lets us immediately inform the user of a path issue.

**Test**:
- CMD+P → “Save DB”, enter `current_run`, check `<workspace>/current_run`.

---

### From Elodin CLI ✔️
Needs two terminals:
- Run a simulation example in Terminal 1:  
  `cargo run --manifest-path=apps/elodin/Cargo.toml editor examples/drone/main.py`
- In Terminal 2, run: `cargo run -p elodin-db -- lua`
- In Terminal 2, run: `client = connect("127.0.0.1:2240")`
- In Terminal 2, run: `client:save_archive("current_run", "native")`
- Check the DB archive at `<workspace>/current_run`

---

### From Python ✔️
- Open a terminal:
```bash
cd libs/nox-py
uv venv
source .venv/bin/activate                # if not already active
uvx maturin develop --uv                 # rebuild with the new code

```
- Open a Python interpretor:
```python
from examples.ball.sim import world, system
w = world()
exec = w.build(system())
exec.run(600)
exec.save_archive("current_run2", "native")
```
- Check the DB archive at `<workspace>/current_run2`

## Extra Paths & Directories tests
| Path                        | Log                                                                              | Save attempt from UI |
|-----------------------------|----------------------------------------------------------------------------------|----|
| `snapshots/run_2025_03_01`  | `INFO Saved db to host:/cwd/snapshots/run_2025_03_01`                       | ✅ |
| `/Users/alice/Desktop/run1` | `INFO Saved db to host:/Users/alice/Desktop/run1`                           | ✅ |
| `/mnt/c/Users/alice/run2`   | `ERROR Cannot save db to host:/mnt/c/Users/alice/run2, No such file or directory` | ❌ |
| `C:\Users\alice\run3`       | `ERROR Cannot save db to host:C:\Users\alice\run3\, the path looks like a Windows-style location...` | ❌ |
| `../other_workspace/run4`   | Request blocked by UI validation; no log emitted                                | ❌ |
| `\\nas01\share\project_db`  | Same handling as the `C:\...` path on non-Windows DB hosts                       | ❌ |

### DIFF
--- a/libs/db/src/lib.rs
+++ b/libs/db/src/lib.rs
@@ -26,7 +26,7 @@ use std::{
     collections::HashMap,
     ffi::OsStr,
     fs::{self, File},
-    io::{self, Read, Seek, SeekFrom, Write},
+    io::{self},
     net::{SocketAddr, ToSocketAddrs},
     ops::Range,
     path::{Path, PathBuf},


anderspitman:
@zxq82lm @shanecelis still pretty early in my review but have some questions that can be answered while I'm going through it. First off, with the new barrier stuff in mind, have we done any benchmarking to verify write performance isn't degraded?

x46085:
> @zxq82lm @shanecelis still pretty early in my review but have some questions that can be answered while I'm going through it. First off, with the new barrier stuff in mind, have we done any benchmarking to verify write performance isn't degraded?

If useful knowledge, all the CI example tests are benchmark tests, you could compare say the cubesat test from this PR:
https://buildkite.com/elodin/elodin/builds/1115#0199e87b-2f56-449c-b45e-e483e059c44b

With a previous run without this PR:
https://buildkite.com/elodin/elodin/builds/1117#0199e8c2-ee2f-4e2f-85f8-c5ff2a83fa5c

anderspitman:
Nice. Is `real_time_factor` what we're looking for or am I missing a more obvious value?

shanecelis:
The maximum latency would be however long it takes to copy the DB files. The files could be large, but I don't know off-hand how large. But the performance aspect is less worrisome to me than the soundness of making the copy. In fact write performance being degraded would be a sign that what we're doing is working. 

# Test Write Barrier Idea

If instead of copying the DB, we could put in a busy loop `loop {}`, you'd hope to see the DB no longer being written to. So if you ran a sim, asked the DB to copy—which will hang the copying db's thread—then you'd hope to see to further information being written to the database. One would hope the DB would buffer any further writes but after a while it should fail in some way.

anderspitman:
Sorry, I wasn't very clear. Definitely write performance should be affected when saving the DB. I was more wondering if the implementation is going to affect performance during normal operation, due to the cost of checking for a write lock. I wouldn't even worry about it (premature optimization) except the docs make it very clear the DB and impeller were designed to be extremely performance sensitive.

libs/db/src/lib.rs
```diff
@@ -134,6 +384,89 @@ impl DB {
...
+        let db_state_path = self.path.join("db_state");
+        if db_state_path.exists() {
+            File::open(&db_state_path)?.sync_all()?;
+        }
+        File::open(&self.path)?.sync_all()?;
+        Ok(())
+    }
+
+    pub fn copy_native(&self, target_db_path: impl AsRef<Path>) -> Result<PathBuf, Error> {
+        let target_db_path = target_db_path.as_ref();
```
anderspitman: Almost all the complexity in this function is devoted to preventing the user from writing to certain directories. Do we really need to protect them from all these permutations? I'm all for protecting the user from likely mistakes, but a lot of these seem relatively unlikely. I'm probably missing some context. Maybe these protections were specifically requested, or they are more dangerous than I realize?

libs/db/src/lib.rs
```diff
@@ -58,9 +60,256 @@ mod msg_log;
...
+        } else {
+            dst_file.write_all(&buf[..read])?;
+        }
+    }
+    dst_file.set_len(metadata.len())?;
+    dst_file.sync_all()
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn try_reflink(src: &Path, dst: &Path) -> io::Result<()> {
```
anderspitman: Are there any rust libraries that could handle cross-platform reflink-or-copy funtionality? Maybe not worth it since the code is already written, but we might at least consider breaking this into a separate crate. Seems like a useful feature and maybe others would use it/bang on it as well.

zxq82lm:
> Sorry, I wasn't very clear. Definitely write performance should be affected when saving the DB. I was more wondering if the implementation is going to affect performance during normal operation, due to the cost of checking for a write lock. I wouldn't even worry about it (premature optimization) except the docs make it very clear the DB and impeller were designed to be extremely performance sensitive.

in normal operation the only overhead I see is bumping an `active_writers` counter on each write. We keep that counter so that, when a snapshot starts, we can let in-flight writes drain cleanly. The barrier (`mutex/condvar`) only engages during a snapshot, so steady-state performance shouldn’t be affected.
Thanks for flagging this: I hadn’t specifically optimized that path. Do you think the increment could be done more efficiently? If it ever shows up in profiles, we could maybe switch to an atomic fast-path (to investigate further).
Really appreciate the careful eye, thank you!

libs/db/src/lib.rs
```diff
@@ -58,9 +60,256 @@ mod msg_log;
...
+impl SnapshotBarrier {
+    pub fn new() -> Self {
+        Self {
+            state: StdMutex::new(SnapshotState::default()),
+            cv: Condvar::new(),
+        }
+    }
+
+    pub fn enter_writer(&self) -> SnapshotWriterGuard<'_> {
+        let mut state = self.state.lock().unwrap();
```
anderspitman: This might be some rust ignorance showing but is it generally accepted to unwrap on mutexes rather than having `enter_writer` et al return a `Result`?

libs/db/src/lib.rs
```diff
@@ -1015,22 +1360,28 @@ async fn handle_packet<A: AsyncWrite + 'static>(
             tx.send_msg(&ArrowIPC { batch: None }).await?;
         }
         Packet::Msg(m) if m.id == SetMsgMetadata::ID => {
+            let _snapshot_guard = db.snapshot_barrier.enter_writer();
```
anderspitman, line 1280: AI claims you also need one of these for `SetComponentMetaData`

libs/db/src/lib.rs
```diff
@@ -58,9 +60,256 @@ mod msg_log;
 pub(crate) mod time_series;
 mod vtable_stream;
 
+pub struct SnapshotBarrier {
```
anderspitman, line 63: There's really no way for me to verify the correctness of the locking code by inspection, but I do have one question about the design. It seems dependent on correct usage, ie we need to manually include calls to `enter_writer` everywhere we need protection. Is there any way to instead wrap the protected resource in some sort of mutex or other concurrency primitive, so it's impossible to make writes without acquiring a lock?

As an example, GPT5 claims that we need a guard for `SetComponentMetadata`. I have no idea if that's true, but it becomes a non-issue if the protected resource is wrapped in a mutex.

Unfortunately if such a design is even possible here, I suspect it might be way more invasive, or have performance implications.

anderspitman:
@zxq82lm @shanecelis review submitted. I won't be able to work on this again until tomorrow other than answering questions probably.

anderspitman:
> > Sorry, I wasn't very clear. Definitely write performance should be affected when saving the DB. I was more wondering if the implementation is going to affect performance during normal operation, due to the cost of checking for a write lock. I wouldn't even worry about it (premature optimization) except the docs make it very clear the DB and impeller were designed to be extremely performance sensitive.
> 
> in normal operation the only overhead I see is bumping an `active_writers` counter on each write. We keep that counter so that, when a snapshot starts, we can let in-flight writes drain cleanly. The barrier (`mutex/condvar`) only engages during a snapshot, so steady-state performance shouldn’t be affected. Thanks for flagging this: I hadn’t specifically optimized that path. Do you think the increment could be done more efficiently? If it ever shows up in profiles, we could maybe switch to an atomic fast-path (to investigate further). Really appreciate the careful eye, thank you!

Hm not sure. Honestly I doubt the overhead is very much since it will almost never be locked.

libs/db/src/lib.rs
zxq82lm: In fact, we could very well have no checks on the backend. We do have the option to enforce checks on the target directory on the backend side, which would apply to all clients (Elodin CLI, Python, UI). 
Since we can also add additional checks on the client side (UI only), I added a few there as well, and their existence is also questionable. 
I don’t really have a firm opinion on this; it’s mostly a UX question.

libs/db/src/lib.rs
zxq82lm: Great idea. We could lean on reflink-copy (see [reflink_copy::reflink_or_copy](https://docs.rs/reflink-copy/latest/reflink_copy/fn.reflink_or_copy.html))?

libs/db/src/lib.rs
shanecelis: Good question. It is generally accepted to [unwrap locks](https://users.rust-lang.org/t/should-i-unwrap-a-mutex-lock/61519). It's a fail-fast strategy. It should only happen if a thread panics while holding a lock thereby poisoning it, or the same thread tries to acquire a lock it already holds. I would categorize those panics as programmer errors rather than runtime errors. Runtime errors the code has to deal with because they may happen; programmer errors we programmers have to deal with to ensure they don't happen.

libs/db/src/lib.rs
```diff
@@ -58,9 +60,256 @@ mod msg_log;
...
+        let entry = entry?;
+        let file_type = entry.file_type()?;
+        let src_path = entry.path();
+        let dst_path = dst.join(entry.file_name());
+        if file_type.is_dir() {
+            copy_dir_native(&src_path, &dst_path)?;
+        } else if file_type.is_file() {
+            copy_file_native(&src_path, &dst_path)?;
+        } else {
+            // skip non-regular entries
```
shanecelis, line 222: Maybe we can log what we skipped here.

libs/db/src/lib.rs
```diff
@@ -134,6 +384,89 @@ impl DB {
...
+
+        if run_dir
+            .file_name()
+            .and_then(|n| n.to_str())
+            .map(|n| n.eq_ignore_ascii_case("db"))
+            .unwrap_or(false)
+        {
+            return Err(Error::Io(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "directory name \"db\" is not allowed",
```
shanecelis: What's an example of a path that is invalid because of having the "db" name?

libs/db/src/lib.rs
```diff
@@ -221,6 +555,7 @@ impl DB {
 
     pub fn insert_vtable(&self, vtable: VTableMsg) -> Result<(), Error> {
         info!(id = ?vtable.id, "inserting vtable");
+        let _snapshot_guard = self.snapshot_barrier.enter_writer();
         self.with_state_mut(|state| {
```
shanecelis, line 475: Is it possible or desirable to put the barrier within the `with_state_mut()`?

libs/db/src/lib.rs
```diff
@@ -240,6 +575,7 @@ impl DB {
     }
 
     pub fn push_msg(&self, timestamp: Timestamp, id: PacketId, msg: &[u8]) -> Result<(), Error> {
+        let _snapshot_guard = self.snapshot_barrier.enter_writer();
         let exists = self.with_state(|s| {
```
shanecelis, line 495: In general a `with_state()` no mut, wouldn't need a barrier, would it? I see there's a `with_state_mut()` down below that could require it.

libs/db/src/lib.rs
anderspitman: I don't feel strongly about this either, other than generally trying to keep the code minimal. @shanecelis?

libs/db/src/lib.rs
anderspitman: It's a tough call. That library doesn't seem to be very heavily used. Honestly I say we keep your implementation as long as it's working. If we find a serious bug or something in the future we can always switch to a dependency.

shanecelis: I say let's use reflink-copy. I actually didn't know about this reflink feature, and it worried me when dealing with currently open file handles, but google's AI says what I want to hear, but unfortunately its links only go to more AI. I'd like to find an authoritative source to confirm it.

> macOS's APFS (Apple File System) supports copy-on-write (CoW) functionality, which means files can be reflinked or cloned. This allows for efficient copies where the new file initially shares the same data blocks as the original, only diverging when modifications occur.

> When a file is open and a reflinked copy is made, the behavior regarding subsequent modifications depends on whether the original file or the reflinked copy is being modified:

> Modifying the Original File: If the original file, while open, is modified, APFS's copy-on-write mechanism ensures that the reflinked copy remains unchanged. When the original file is written to, APFS allocates new blocks for the modified data, leaving the shared blocks intact for the reflinked copy.

> Modifying the Reflinked Copy: Similarly, if the reflinked copy is modified while the original file is open, the same CoW principle applies. New blocks are allocated for the modified data in the reflinked copy, and the original file's shared blocks remain untouched.

libs/db/src/lib.rs
shanecelis: I'm in favor of having checks here.

libs/db/src/msg_log.rs
```diff
@@ -124,6 +125,17 @@ impl MsgLog {
         self.waker.clone()
     }
 
+    pub fn sync_all(&self) -> Result<(), Error> {
+        self.timestamps.sync_all()?;
+        self.bufs.sync_all()?;
+        let metadata_path = self.path.join("metadata");
+        if metadata_path.exists() {
+            File::open(&metadata_path)?.sync_all()?;
```
shanecelis, line 133: I don't quite understand what we're doing with the metadata file here. We open it, if it's not already open? I'm guessing that's a non-mem-mapped file. Are we just ensuring it's up to date?

libs/db/src/lib.rs
zxq82lm: Here, if  the mutex is poisoned, the process is already in a bad state (some other thread panicked while holding the lock). We don’t have a sensible recovery strategy for that situation; after a panic inside the snapshot barrier the server is effectively inconsistent. In that kind of code it’s common to fail fast rather than propagate a Result that every caller would just unwrap anyway. To make that intention explicit I changed the `unwrap()` for:
```rust
let mut state = self
    .state
    .lock()
    .expect("snapshot barrier mutex poisoned");
```

anderspitman: Makes sense, thanks for the explanation!

libs/db/src/lib.rs
zxq82lm: You are right: this query does write to disk (`ComponentMetadata::write`), so it needs to be guarded by the barrier to prevent any race conditions during `save_archive`.

libs/db/src/lib.rs
zxq82lm: I like the idea of treating `reflink-or-copy` as a standalone feature. I can explore extracting it into a new crate and adding a README. What do you think @shanecelis?

libs/db/src/lib.rs
zxq82lm: In theory you could wrap the whole subsystem in a mutex-like primitive, but that would be a sizable architectural change: we’d have to route every DB entry point through the guard, it would ripple across the entire stack, and you’d pay in contention and latency. Thanks for catching the missing guard around SetComponentMetadata; that was indeed an oversight on our side, and now we cover that call with the existing protection.

libs/db/src/lib.rs
shanecelis: No, let's not do a new crate. I think we should use the [reflink_copy](https://crates.io/crates/reflink-copy) crate that already exists or do a plain recursive copy.

anderspitman (approved):
@zxq82lm just noticed one weird thing, and I'm not sure how important it is because it will only affect users running the DB on a different machine than the editor (or running the simulation in WSL like I am).

Basically the save path appears to be relative to the editor working directory, and the editor resolves the absolute path and sends that to the DB which then saves the archive at that path, but if they are communicating across a network then the path might not exist from the DB's perspective.

Maybe we should just save relative to the DB, and the editor just passes a name?

shanecelis:
That's an important consideration, @anderspitman. Thanks for bringing it up. I like that the current behavior will put it where you requested it _if_ it's on the same machine. If it is remote and that path isn't present, it should fail. Perhaps the thing to do is to log this like so:
```
INFO Saved db to `hostname:/path/to/db/`
```
or
```
ERROR Cannot save db to `hostname:/path/to/db`, no such directory.
```
The only thing I'd want to address, how can I specify a path that doesn't include my current path since the remote host may not have it? I'd suggest that if the user submits an absolute path to save to, then we do not prepend the Current Working Directory (CWD).

anderspitman:
One thing to be aware of is it might succeed in exporting even if the parent path doesn't exist. For example I was running the DB in WSL and the editor native, and when I exported it created a single directory (not a tree) named something like this: 'C:\Users\anders\current_db'. So the db was archived to 'C:\Users\anders\current_db'/db

zxq82lm:
> One thing to be aware of is it might succeed in exporting even if the parent path doesn't exist. For example I was running the DB in WSL and the editor native, and when I exported it created a single directory (not a tree) named something like this: 'C:\Users\anders\current_db'. So the db was archived to 'C:\Users\anders\current_db'/db

---

Nice catch, @anderspitman, thanks a lot! The next commit (which I’m finishing and testing) should now provide this behavior in this exact case (@shanecelis):
- The server (WSL Linux) detects `C:\Users\anders\current_db` as a Windows absolute path and rejects it immediately: it logs `ERROR Cannot save db to wsl-host:C:\Users\anders\current_db\, no such directory`.
- The export therefore fails before any folders are created; no tree or phantom directory is produced on the WSL side.
- The UI receives exactly this error message from the server and shows it to the user, which lets them correct the path. For example, `/mnt/c/Users/anders/current_db`.

Could you please rerun the same test after I’ve pushed this commit? Thank you. The issue is hard for me to reproduce locally.

zxq82lm:
@shanecelis @anderspitman The latest commits should sort out the _directory and path issues_ we hit. Thanks again for the heads-up, please ping me if anything still looks off!

shanecelis:
I am trying to do the following. Run the rocket example, save the DB, and then load that DB.

```sh
cargo run --manifest-path=apps/elodin/Cargo.toml editor examples/rocket/main.py
```
I pause it (but that doesn't stop DB writes) then I save the DB to 'a'. Now when I try to load up DB a, I get this error:

```sh
$ elodin-db run 0.0.0.0:2240 a
2025-10-24 12:21:25.198  INFO starting db path="a"
2025-10-24 12:21:25.198  INFO listening addr=0.0.0.0:2240
2025-10-24 12:21:25.198  INFO Opening database: a
Error:   x invalid component id

```

Now when I try to load the db mentioned in the log from its temp directory, I get a similar error:

```sh
$ elodin-db run 0.0.0.0:2241 /var/folders/wv/4qnq4g1j3kg5gj5q6jzq4s7h0000gn/T/.tmptgpKMk/db
2025-10-24 12:25:49.770  INFO starting db path="/var/folders/wv/4qnq4g1j3kg5gj5q6jzq4s7h0000gn/T/.tmptgpKMk/db"
2025-10-24 12:25:49.770  INFO listening addr=0.0.0.0:2241
2025-10-24 12:25:49.770  INFO Opening database: /var/folders/wv/4qnq4g1j3kg5gj5q6jzq4s7h0000gn/T/.tmptgpKMk/db
Error:   x No such file or directory (os error 2)
  `-> No such file or directory (os error 2)

```

zxq82lm:
> I am trying to do the following. Run the rocket example, save the DB, and then load that DB.
> ```
> Error:   x No such file or directory (os error 2)
>   `-> No such file or directory (os error 2)
> ```

---
Thanks so much for spotting this problem, @shanecelis.
Turns out it was a too-strict LoadDB check:
- a/db/7106446861283339143 is just the identity card for the Globals component. The metadata file literally contains “Globals”, but there’s no schema or data because it’s the singleton we use for global simulation state.
- a/db/3810108550136042869 corresponds to the rocket component, so it only stores that component’s metadata; no schema or data files are expected there.
- During DB load, we validate each component directory. Because these folders only had metadata, the loader was aborting with “No such file or directory”.

I fixed it by making DB::open skip directories that don’t have a schema. The snapshot now loads fine; we just warn when we see metadata-only components:

```
$ cargo run -p elodin-db -- run 0.0.0.0:2240 ./a/db
```

**Output**:
```
2025-10-24 11:06:19.034  INFO starting db path="./a/db"
2025-10-24 11:06:19.034  INFO listening addr=0.0.0.0:2240
2025-10-24 11:06:19.035  INFO Opening database: ./a/db
2025-10-24 11:06:19.038  WARN Skipping component without schema while opening database component.id=3810108550136042869 component.dir=./a/db/3810108550136042869
2025-10-24 11:06:19.047  WARN Skipping component without schema while opening database component.id=7106446861283339143 component.dir=./a/db/7106446861283339143
2025-10-24 11:06:19.055  INFO opened db db.path="./a/db"
```
I’m still running a couple of follow-up tests, but the loader no longer crashes on these snapshots.

shanecelis:
Thanks for the fast fix, Benoît! I confirm that it's working for the case I showed above.

When I give it the path I saved to 'c', it does force me to specify it as 'c/db' on the CLI. On the plus side, it gives a non-zero exit code.

```sh
$ elodin-db run 0.0.0.0:2240 c; echo $?
2025-10-24 15:20:01.068  INFO starting db path="c"
2025-10-24 15:20:01.068  INFO listening addr=0.0.0.0:2240
2025-10-24 15:20:01.068  INFO Opening database: c
Error:   x invalid component id

1
```

I'd like to catch this error where I try to open an existing directory that doesn't have a DB in it and provide a more helpful error message like: "No `c/db_state` file found." That should help the user.

When we provide a non-existent directory, it creates a new database there without any interior 'db' directory. Would it be possible to drop the extra interior 'db' directory?

zxq82lm:
> 1/ I'd like to catch this error where I try to open an existing directory that doesn't have a DB in it and provide a more helpful error message like: "No c/db_state file found." That should help the user.
```
cargo run -p elodin-db -- run 0.0.0.0:2240 ./c  
```
**Output**:
```
2025-10-24 13:08:28.372  INFO starting db path="./c"
2025-10-24 13:08:28.373  INFO listening addr=0.0.0.0:2240
2025-10-24 13:08:28.373  INFO Opening database: ./c
Error:   × missing db_state file at ./c/db_state
```
---
> 2/ When we provide a non-existent directory, it creates a new database there without any interior 'db' directory. Would it be possible to drop the extra interior 'db' directory?
- There are no longer any references to a /db subdirectory.

shanecelis:
I confirm the presence of the error about a missing `db_state` file AND that the interior 'db' directory has been removed from saves. Thank you, Benoît!

libs/db/src/lib.rs
zxq82lm: Subdirectory /db is not used anymore.

libs/db/src/lib.rs
zxq82lm: It is covered by:
warn!("Skipping irregular file for db copy {:?}", src_path.display())

libs/db/src/lib.rs
zxq82lm: We need the snapshot barrier outside with_state_mut: once inside, we only have &mut State, so we can’t access self.snapshot_barrier, and the barrier must cover the entire mutation while the write lock is held.

libs/db/src/lib.rs
zxq82lm: Even though the first with_state is read-only, it’s part of a read–then–maybe-write sequence. We enter the barrier once so the entire operation stays atomic with respect to snapshots: if the log doesn’t exist, we immediately call with_state_mut under the same guard to create it. Skipping the barrier for the read would let a snapshot slip in between the existence check and the subsequent write path, so we keep the guard across both calls.

libs/db/src/msg_log.rs
zxq82lm: It’s just making sure the metadata postcard we wrote earlier actually hits disk: metadata.write() opens the file, writes, and closes it, so the only way to fsync is to reopen briefly and call sync_all(). Since the metadata file isn’t memory-mapped (unlike the append logs), we don’t keep a handle around. So that reopen-and-sync is intentional.

PR #242, opened 2025-10-15
Refactor EQL formulas (docs + tests) & EqlContext bug fixing (#248)

zxq82lm:
## PR Objective
- The main purpose of this PR is to reorganize the eql directory to have a subdirectory containing the EQL formulas, equipped with a README.md that must be automatically integrated into the documentation (its content explains how to add a new EQL formula).

**New organization**:
```
libs/db/eql/src/
├── eql_formulas/
│   ├── fft.rs
│   ├── fftfreq.rs
│   ├── first.rs
│   ├── last.rs
│   ├── mod.rs
│   └── norm.rs
│   ├── README.md # ex-how-to-new-formula.md
└── README.md
└── lib.rs
```
- During this PR, a gap was detected that prevents first() and last() from functioning correctly. They need the simulation’s time range (start, end) provided by a context, which the UI never updates. We take advantage of this PR to propagate the updated context to any formula. See [#248](https://github.com/elodin-sys/elodin/issues/248).

---
---

### Reorganization

**Tests**:
```
cargo test -p eql
```

**Documentation**:
```
cargo doc -p eql --no-deps
cargo doc -p eql --open
```

---

### Bug fixing
See [#248](https://github.com/elodin-sys/elodin/issues/248)

**Tests**:
- Run `cargo run --manifest-path=apps/elodin/Cargo.toml editor examples/ball/main.py`
-  CMD+P > Create Query Table

- Apply these EQL macros (first 1 s, ISO standard, or simplified writing accepted by our parser):
```
ball.world_pos.first("PT1S")
ball.world_pos.first("1sec")
```
```
ball.world_pos.last("PT0.5S")
```

And these ones: (last 5 minutes, ISO standard, or simplified writing accepted by our parser):
```
ball.world_pos.last("PT5m")
ball.world_pos.last("5m")
```

- Test also others available formulas (norm, fftfreq, fft):
```
ball.world_pos.norm()
```
```
ball.world_pos.time.fftfreq()
```
```
ball.world_pos[4].fft().last("PT2S")
```

### DIFF
--- a/libs/db/eql/src/formulas/first.rs
+++ b/libs/db/eql/src/formulas/first.rs
@@ -1,5 +1,0 @@
-use crate::{Context, Error, Expr, parse_duration};
-use std::sync::Arc;
-
-
-#[cfg(test)]


zxq82lm:
See also: https://github.com/elodin-sys/elodin/issues/251, identified during this PR.

libs/db/eql/src/eql_formulas/norm.rs
```diff
@@ -0,0 +1,47 @@
...
+        ));
+    }
+    let n_elems: usize = dims.iter().copied().map(|d| d as usize).product();
+    let mut terms = Vec::with_capacity(n_elems);
+    for i in 0..n_elems {
+        let qi = Expr::ArrayAccess(Box::new(Expr::ComponentPart(part.clone())), i)
+            .to_qualified_field()?;
+        terms.push(format!("{qi} * {qi}"));
+    }
+    Ok(format!("sqrt({})", terms.join(" + ")))
```
shanecelis: "sqrt"?

shanecelis:
This looks good. 

As an aside, one thing that's odd here is where is the FFT implemented? It must be elsewhere. It seems like what we can implement ourselves in EQL are functions that transform arguments into SQL like `norm()` does.

zxq82lm:
> This looks good.
> 
> As an aside, one thing that's odd here is where is the FFT implemented? It must be elsewhere. It seems like what we can implement ourselves in EQL are functions that transform arguments into SQL like `norm()` does.

@shanecelis In this specific case (FFT) we need a UDF because Fourier transforms aren’t available out of the box in the database (unlike `sqrt` for `norm()`). As a result, it’s implemented as the FftUDF in libs/db/src/arrow/fft.rs, which calls `compute_fft_magnitudes` (also in libs/db/src/arrow/fft.rs).

libs/db/eql/src/eql_formulas/norm.rs
zxq82lm: Yes. "sqrt" is available as a native SQL scalar, so we rely on it directly; no UDF is needed.

PR #247, opened 2025-10-16
Update macOS local development instructions

joelreymont:


.cargo/config.toml
joelreymont, line 1: I'm open to putting this in the README as it may, hypothetically, conflict with libraries provided by the Nix development shell.

shanecelis:
Joel, what is your macOS version? Here is mine.
```sh
$ sw_vers
ProductName:		macOS
ProductVersion:		15.6.1
BuildVersion:		24G90
```
I'm just curious why it's necessary for you but not us.

joelreymont:
❯ sw_vers
ProductName:            macOS
ProductVersion:         26.0.1
BuildVersion:           25A362

shanecelis:
SKINNER: Am I so out of touch? 

Skinner-Out-Of-Touch

SKINNER: No, it's the updates who are waiting.

Screenshot 2025-10-16 at 8 41 39 AM

Thank you. That explains why.

joelreymont:
I don't think it explains it but I'm on it!

joelreymont:
We may want to add a note that if `-liconv` is missing then you are building with `gcc` and should use `clang` instead.

shanecelis:
Closed?

joelreymont:
> Closed?

Not needed. Only if you build using gcc which you should not. Using clang works fine using our current instructions.

PR #249, opened 2025-10-16
Consolidate examples

### LINKED ISSUE #83: Consolidate examples
The current Elodin Sim examples include a directory at the root, and a directory in libs/nox-py/examples, which typically goes undiscovered. Since the current tutorial references the bouncing ball template in the root examples folder, suggest that all examples consolidate there. Bonus to update the tutorial documentation to reference the many examples to increase discoverability.

### LINKED ISSUE #143: Create READMEs for each example in libs/nox-py/examples
Proposed structure:
```
examples/
├── cube-sat/
│   ├── cube-sat.py
│   └── cube-sat-pysim.py
│   └── README.md
├── rocket/
│   ├── rocket.py
│   └── README.md
├── three-body/
│   ├── three-body.py
│   └── README.md
```

Each README can briefly describe:
- what the example does,
- required dependencies,
- how to run it

anderspitman:
Move examples from `libs/nox-py/examples/` into their own directories under `examples/`. Also added basic README files.

Fixes #83 

Fixes #143 

I'm using a slightly different structure to what was proposed in #143, where I used for example `examples/three-body/main.py` instead of `examples/three-body/three-body.py`. I don't have strong feelings about this and can change it if preferred. @x46085 @shanecelis thoughts?


anderspitman:
> I'm afraid you're going to have to search for "../../examples" and probably fix the many references that exist in the docs.

Can you be more specific about the problem you're seeing? AFAIK all the references in the docs have been updated and are accurate.

shanecelis:
Sure. Here's one:
```diff
diff --git a/apps/elodin/README.md b/apps/elodin/README.md
index 433ad29a..53267165 100644
--- a/apps/elodin/README.md
+++ b/apps/elodin/README.md
@@ -7,9 +7,6 @@ Install the editor from the [releases](https://github.com/elodin-sys/elodin/rele
 To run the editor locally and test the included examples, follow these steps:
 
 ```bash
-# Move into the nox-py library
-cd libs/nox-py
-
 # Create a new virtual environment using uv
 uv venv
 
@@ -20,7 +17,7 @@ source .venv/bin/activate
 uvx maturin develop --uv
 
 # Run the three-body example from the editor (execute inside libs/nox-py)
-cargo run --manifest-path=../../apps/elodin/Cargo.toml editor ../../examples/three-body/main.py
+cargo run --manifest-path=apps/elodin/Cargo.toml editor examples/three-body/main.py
```

But if you do a recursive search for "../../examples" you'll find more.

anderspitman:
I believe that command is correct. Note that the comment above it says to run it from inside the `libs/nox-py` directory. Obviously we can change that if we want but I tried to keep everything as close to the way it currently works while implementing the change.

anderspitman:
Or is that what you're saying, that you want to change all the examples to be run from the same working directory, ie the top level?

x46085:
> Or is that what you're saying, that you want to change all the examples to be run from the same working directory, ie the top level?

Adjusting all instructions to operate from top level sounds like a positive change. (make sure you pull in the latest).

shanecelis:
Yes, let's try to run from the root. 

Oh, I see, you had updated it for the new location. Apologies for not specifying this desire earlier.

I suppose references to "../../apps" would also be things to correct.

x46085 (approved):

anderspitman:
I agree we should focus on top level. I think the only place it makes sense to have the `../../` is in the docs for nox-py itself, since it's likely the user will be doing development out of that directory.

anderspitman:
Ok I looked at the places where we're currently using "../../". It's not ideal, but I think it is at least correct, ie the instructions work. I don't think it's worth improving in this PR because I'm going to be doing an example overhaul PR soon which will completely get rid of the `elodin editor` command, and I'll need to update all these anyway from that perspective, and might be able to make additional improvements with the context I'll have then.

x46085:
Agree, fire away.

Issue #243, opened 2025-10-15
"Command Palette > Set time range" panics on invalid start/end bounds


zxq82lm:
### Summary

With the ball example: setting the time range via the palette to `start = +10s` / `end = -0s` makes the editor panic:
```
thread 'Compute Task Pool (4)' panicked at ...nodit.../utils.rs:255:3:
invalid interval given to function see here for more details: https://docs.rs/nodit/latest/nodit/#invalid-intervals
Encountered a panic in system elodin_editor::ui::plot_3d::gpu::extract_lines!
```

The interval is absurd (start after the dataset, end 0 s before the latest point) but it should be rejected, not crash the whole app.

### Steps to Reproduce

1. Launch the `ball` example (`cargo run --manifest-path=apps/elodin/Cargo.toml -- run examples/ball/main.py`).
3. CMP+P → **Set Time Range**  
   * Start offset: `+10s`  
   * End offset: `-0s`
4. Submit

### What happens

The `extract_lines` system panics because `nodit` receives `start > end`, propagating a panic and killing the editor.

### What should happen instead

The UI should validate the time range before applying it. If `start ≥ end`, show a user-facing error (“invalid time range”) and keep the previous range. At minimum, guard against the empty interval in `TimeRangeBehavior::calculate_selected_range` before sending it down to nodit.

PR #252, opened 2025-10-17
Validate command-palette time range input and guard runtime selection

zxq82lm:
## PR Objective
This PR addresses issue [#243](https://github.com/elodin-sys/elodin/issues/243), related to Command Palette > Set Time Range.

## Description
Hardened time-range handling so `TimeRangeBehavior::calculate_selected_range` rejects empty or inverted intervals, keeps the previous selection when no data is present, and logs a warning instead of panicking.

## Unit tests
```bash
cargo test -p elodin-editor time_range_tests
```


shanecelis (approved):
I didn't even know about this feature, but I was able to exercise it without issue.

PR #259, opened 2025-10-22
Use an EqlFormula trait if possible

shanecelis:
Hi Benoît,

Your modularization is a great start. However, it looks like it'll require some hand massaging to integrate each new formula into the lib. My hope is that it's a zero, one, or two line change to lib. It doesn't have to be fully dynamic. It can be decided at compile time.

There's also what seem to be the beginnings of a contract with the `parse` function. I have tried to sketch out what the usage of the trait `EqlFormula` to enforce and enable that contract might look like starting with the fft function. I don't believe it's comprehensive though. It seems different formulas have different methods for reasons that aren't entirely clear to me, but I'd like for us to refine `EqlFormula` to suit any given function.

Take a look and we'll discuss.

### DIFF
--- a/libs/db/eql/src/eql_formulas/norm.rs
+++ b/libs/db/eql/src/eql_formulas/norm.rs
@@ -67,12 +67,11 @@ impl super::EqlFormula for Norm {
     }
 
     fn suggestions(&self, expr: &Expr, _context: &Context) -> Vec<String> {
-        if let Expr::ComponentPart(part) = expr {
-            if let Some(component) = &part.component {
-                if !component.schema.dim().is_empty() {
-                    return vec!["norm()".to_string()];
-                }
-            }
+        if let Expr::ComponentPart(part) = expr
+            && let Some(component) = &part.component
+            && !component.schema.dim().is_empty()
+        {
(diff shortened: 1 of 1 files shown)


zxq82lm:
> It seems different formulas have different methods for reasons that aren't entirely clear to me, but I'd like for us to refine `EqlFormula` to suit any given function.

Each formula exposes bespoke helpers because their behaviours differ: 
- `FFT`/`FFTFREQ` need field and column munging tightly coupled to their DataFusion UDFs.
- `First`/`Last` wrap SQL with time-window clauses and custom completion hints.
- `Norm` validates vector dimensions and augments suggestions.

---

@shanecelis: We can tighten `EqlFormula` into a single contract (parse, SQL rendering, contextual suggestions, with opt-outs where they don’t apply), back it with a static registry the parser can lean on, and actually route calls through `Expr::Formula`. Once that scaffolding is in place we can migrate `FFT`, `FFTFREQ`, `First`, `Last`, and `Norm` to the shared path, delete the redundant helpers, and refresh the tests to cover parsing, SQL output, and suggestions end to end.

PR #267, opened 2025-10-27
fix: Avoid subtract overflow.

shanecelis:
When I ran the rocket example once, I got this:
```
thread 'main' panicked at libs/elodin-editor/src/ui/plot/widget.rs:1204:13:
attempt to subtract with overflow
```
It was not reproducible.

p.s. This PR also has a one-line change to the justfile from doing the release.

### DIFF
--- a/justfile
+++ b/justfile
@@ -23,6 +23,7 @@ promote tag:
   #!/usr/bin/env sh
   dir=$(mktemp -d)
   gh release download {{tag}} --pattern 'elodin-*.whl' --dir $dir
+  echo "uv publish \"$dir/*.whl\""
   uv publish "$dir/*.whl"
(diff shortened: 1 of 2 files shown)


x46085 (approved):

zxq82lm (approved):
Pulled the branch locally, it compiles without any issues, and the rocket example runs fine. The fix looks coherent and consistent with the existing code.

shanecelis:
This bug hit me in PR #277. It seems like it and many other numerical overflows are expressed when the db is empty. Because of this I cherry-picked this commit and fixed numerous other overflows there. Therefore, I advocate that we close this PR in favor of #227 which will include this fix and more.

PR #269, opened 2025-10-27
feat: Allow conditional systems.

shanecelis:
We want to express conditional systems in Python at run time. If you have Python systems `a | b | c`  but `b` evaluates to `None` then it ought to behave as though you wrote `a | c`.

### DIFF
--- a/libs/nox-py/src/system.rs
+++ b/libs/nox-py/src/system.rs
@@ -284,7 +284,7 @@ impl System {
         }
     }
 
-    pub fn __ror__(&self, other: PyObject) -> System {
+    pub fn __ror__(&self, _other: PyObject) -> System {
         // Handle the case where the left operand is None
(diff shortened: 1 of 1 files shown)


x46085:
Cool, where did you discover this use case?

x46085 (approved):

shanecelis:
This was straight from Nat and Sung's "How do I do conditional systems?" query re-interpreted as a wish.

x46085:
> This was straight from Nat and Sung's "How do I do conditional systems?" query re-interpreted as a wish.

Right but important to ask the "5 whys" on how they got into a situation that warranted it. I have no issue with the idea, I'm just trying to imagine the situation why it's a must have.

shanecelis:
That's a good point. Alternatively, you could make a dummy no-op system, and do `a if flag else noop_sys`.

PR #280, opened 2025-11-03
Switch some bevy dependencies to upstream sources

anderspitman:
We were using forked versions in order to get bevy 0.16 support, but that's now been upstreamed.


anderspitman:
Yeah it looks ok to me

PR #282, opened 2025-11-04
Fix: axes widget matches world

### LINKED ISSUE #273: Axes widget ought to represent the world axes—not the camera axes
# Problem

The axes widget represents the camera's orientation. We'd prefer it to represent the world's orientation. 

No matter what direction the camera is looking, its initial orientation looks like this:

Image

The camera and the world's orientation will coincide when it's looking West (-x direction), i.e., when it's looking _from_ the East (toward the origin from positive x).

## History

In PR #253 the axes was changed to try and portrait the world axes with respect to the new compass mesh. If one is looking at an object from the East (positive x-axis), it looks like we had achieved that.

# Solution

Draw the axes such that the true world orientation is represented. Red pole points East; Green pole points North; and Blue pole points Up. RGB -> XYZ. This is the ENU coordinate system.

# Acceptance Criteria

When you run the rocket example, and change the camera's starting position, it will always show a consistent relationship between the compass's East and North with the axes' Red and Green poles respectively like so:

Image
 
```sh
cd elodin
cargo run --features elodin-editor/inspector --manifest-path=apps/elodin/Cargo.toml editor examples/rocket/main.py
```

shanecelis:
This PR fixes #273. That issue specifies the acceptance criteria. 

## Old

Here is the old behavior. 
old

Notice the axes widget is shows essentially an identity orientation where green (north) is pointed to the right, while the world axis line given by the drone is saying north is to the right and down.

## New

The new behavior has the axes widget represent the world axes: East North Up (ENU) as Red Green Blue (RGB).

new

### DIFF
--- a/libs/elodin-editor/src/plugins/navigation_gizmo.rs
+++ b/libs/elodin-editor/src/plugins/navigation_gizmo.rs
@@ -251,16 +251,6 @@ pub fn spawn_gizmo(
                 clear_color: ClearColorConfig::None,
                 ..Default::default()
             },
-            // I played with a orthographic projection but it left a lot to be
-            // desired.
-            //
-            // Projection::from(OrthographicProjection {
-            //     // 6 world units per pixel of window height.
-            //     scaling_mode: bevy::render::camera::ScalingMode::FixedVertical {
-            //         viewport_height: 6.0,
-            //     },
-            //     ..OrthographicProjection::default_3d()
-            // }),
             Camera3d::default(),
             render_layers.clone(),
             NavGizmoParent { main_camera },


libs/elodin-editor/src/plugins/navigation_gizmo.rs
```diff
@@ -245,14 +246,26 @@ pub fn spawn_gizmo(
...
+            // I played with a orthographic projection but it left a lot to be
+            // desired.
+            //
+            // Projection::from(OrthographicProjection {
+            //     // 6 world units per pixel of window height.
+            //     scaling_mode: bevy::render::camera::ScalingMode::FixedVertical {
+            //         viewport_height: 6.0,
+            //     },
+            //     ..OrthographicProjection::default_3d()
+            // }),
```
x46085: do you plan to dive right back into experimenting with this? if no, then better to remove and leave clean

shanecelis: Nope. Excising it.

Issue #290, opened 2025-11-05
[Parent issue] Multi-Window Follow Up Features


shanecelis:
## Original plan
* `window screen=1` 
* blank canvas
  * create a panel
  * drag panel to create secondary window
  * position windows on different monitors full screen
* lock serializes to KDL
* plus on secondary windows
* `window fullscreen=#true ` 
* save schematic 

---

## Definitions
- Monitor: the Elodin Monitor object.
- Screen: a physical display.
- Window: a graphical window (here, the main or a secondary one).

---
## Development plan

PR [#293](https://github.com/elodin-sys/elodin/pull/293) (closed)
- 1. ~~Drop {} wrappers from window path KDL definitions and loading.~~
- 2. ~~Make secondary windows editable so the “+” add-pane button is available.~~
- 3. ~~Add "Command Palette > Create Window" to open a blank secondary window.~~
---

PR [#302](https://github.com/elodin-sys/elodin/pull/302)  (opened)
- 4. Extend KDL save/load to persist per-window metadata: physical screen, window position, window size (x, y), and fullscreen state. A basis to elaborate on:
```
window path="examples/drone/rate-control-panel.kdl" screen=1 {
    rect 37 26 33 44    # rect <left %> <top %> <width %> <height %>
}
window path="examples/drone/motor-panel.kdl" screen=2 {
   rect 0 0 100 100     # full screen
}
window screen=0  { # no path argument => main window
   rect 0 0 100 100   
}
```
---
- 5.  ~~Make the current lock shared across all windows.~~  Done by PR [#298](https://github.com/elodin-sys/elodin/pull/298)  (closed)
- 6. Lock serializes to KDL. Sub-issue [#292](https://github.com/elodin-sys/elodin/issues/292)
---
Issue [#291](https://github.com/elodin-sys/elodin/issues/291)
- 7. Allow creating a secondary window that hosts a panel originating from the main window (for example via drag-out).
- 8. Allow the inverse operation: docking a panel from a secondary window back into the main window.

shanecelis:
I'd say "dragging" a panel to create a window is cool, but I fear it may be tricky. If they can "Command-P Create Window" and hit the plus button in a blank window, that is sufficient to enable all uses. So I would not prioritize "dragging"; let's put that into a new ticket #291.

PR #302, opened 2025-11-07
Multi-window: window metadata (physical screen, position, dimension) persistence 

zxq82lm:
## PR objective
The purpose of this PR is to address item 4 from issue [#290](https://github.com/elodin-sys/elodin/issues/290):
- 4. Extend KDL save/load to persist per-window metadata: physical screen, window position, window size (x, y), and fullscreen state.

## Definitions
- Monitor: the Elodin Monitor object.
- Screen: a physical display.
- Window: a graphical window (here, the main or a secondary one).

## Drone example
```  
        window path="examples/drone/rate-control-panel.kdl" screen=2 {
           rect 5 5 40 40  
        }
        window path="examples/drone/motor-panel.kdl" screen=0 {
           rect 0 0 100 100
        }
        window screen=1 {  
           rect 0 0 50 50    
        }
```   

**Explanations**:
```
- rect 5 5 40 40   # rect <left %> <top %> <width %> <height %>
- rect 0 0 100 100   # full screen
- window screen=1   # no path argument => main window
```

>On macOS only: sometimes the OS has already finished leaving fullscreen by the time we call set_fullscreen(None), and AppKit logs a message like:
> ```2025-11-08 16:43:26.588 elodin[15943:3269063] not in fullscreen state```
> This log is harmless and can be safely ignored.


anderspitman (commented):
This is quite a large amount of code to review. Naively looking at the desired feature, I would hope it ended up being a relatively straight forward modification to our ser/de logic. Obviously it turned out trickier than that.

@zxq82lm would you mind writing (or maybe you already have and I missed it) a high level overview of the issues you ran into and your approach to solving them? That would greatly help me fit everything into place while reviewing.

On another note, some of these files are getting quite large. For example, `libs/elodin-editor/src/ui/mod.rs` has nearly quadrupled in size in the last couple months. Would it be possible to break some of this out into smaller files? Ideally, if code can be encapsulated into individual files with self-contained, pluggable functionality I think it would be easier to review and maintain.

@shanecelis do we already have any sort of convention on file size? I don't tend to be too strict with myself about it, but I've found ~150-300 lines to be ideal, <500 pretty good, and <1000 not too bad.

zxq82lm:
> This is quite a large amount of code to review. Naively looking at the desired feature, I would hope it ended up being a relatively straight forward modification to our ser/de logic. Obviously it turned out trickier than that.
> 
> @zxq82lm would you mind writing (or maybe you already have and I missed it) a high level overview of the issues you ran into and your approach to solving them? That would greatly help me fit everything into place while reviewing.
> 
> On another note, some of these files are getting quite large. For example, `libs/elodin-editor/src/ui/mod.rs` has nearly quadrupled in size in the last couple months. Would it be possible to break some of this out into smaller files? Ideally, if code can be encapsulated into individual files with self-contained, pluggable functionality I think it would be easier to review and maintain.
> 
> @shanecelis do we already have any sort of convention on file size? I don't tend to be too strict with myself about it, but I've found ~150-300 lines to be ideal, <500 pretty good, and <1000 not too bad.
---
- This PR hardens multi-window behavior and layout persistence: on Linux we force Xwayland to avoid Wayland quirks (forced fullscreen, ignored geometry); on macOS we wrap fullscreen/maximize transitions with timeouts so we don’t get stuck when the WM drags its feet. We clearly distinguish the primary window from secondary windows, each with its own NeedScreen/NeedRect phase, geometry capture, and render order.
- The KDL screen/rect values are now persisted and locked so that runtime captures don’t overwrite them; secondary schematics are reloaded / de-looped cleanly by ignoring the paths we just wrote.
- Yes, ui/mod.rs has grown because all the multi-window machinery is intertwined (system ordering, relayout states, timeouts, Winit helpers). Refactoring it into several modules (layout, capture, persistence, primary vs secondary) would definitely be nice.
- Doing that after this PR lands feels less risky, so we don’t mix structural refactors and functional fixes in the same change. My suggestion: merge the fixes as-is, then do a follow-up PR that refactors ui/mod.rs into dedicated modules (layout/relayout, geometry capture, persistence, primary vs secondary) so we don’t mix structural cleanup with the current functional changes.

anderspitman:
@zxq82lm thanks for the rundown. I'm doing some generic testing now before looking closer at the code. Overall the multi-window stuff is awesome. Great work!

I am seeing some interesting behavior. If do the following steps:

1. Load drone example
2. Move windows
3. Save schematic
4. Close
5. Load drone example again
6. Save *again*

Elodin goes into some sort of a loop. I'm using i3wm, which is pretty nonstandard, so that might be the problem. I'm curious if you see the same behavior on Ubuntu?

Here's what I'm see in the logs during the loop:

```
2025-11-19 09:40:37.269  INFO Applying primary window rect rect=WindowRect { x: 22, y: 15, width: 57, height: 68 } width_px=2189 height_px=1469 x=845 y=324
2025-11-19 09:40:37.270  INFO Confirmed screen assignment (sync) screen=0 path=motor-panel.kdl
2025-11-19 09:40:37.270  INFO Confirmed screen assignment (sync) screen=0 path=rate-control-panel.kdl
2025-11-19 09:40:37.273  INFO refreshing schematic path="drone.kdl"
2025-11-19 09:40:37.352  INFO Loaded secondary schematic path=motor-panel.kdl
2025-11-19 09:40:37.353  INFO Loaded secondary schematic path=rate-control-panel.kdl
2025-11-19 09:40:37.355  INFO Primary window confirmed on target screen screen=0
2025-11-19 09:40:37.374  INFO refreshing schematic path="drone.kdl"
2025-11-19 09:40:37.438  INFO Loaded secondary schematic path=motor-panel.kdl
2025-11-19 09:40:37.439  INFO Loaded secondary schematic path=rate-control-panel.kdl
2025-11-19 09:40:37.442  INFO Primary window confirmed on target screen screen=0
2025-11-19 09:40:37.476  INFO refreshing schematic path="drone.kdl"
2025-11-19 09:40:37.577  INFO refreshing schematic path="drone.kdl"
2025-11-19 09:40:37.603  INFO Loaded secondary schematic path=motor-panel.kdl
2025-11-19 09:40:37.604  INFO Loaded secondary schematic path=rate-control-panel.kdl
2025-11-19 09:40:37.615  INFO Primary window confirmed on target screen screen=0
2025-11-19 09:40:37.681  INFO Loaded secondary schematic path=motor-panel.kdl
2025-11-19 09:40:37.681  INFO Loaded secondary schematic path=rate-control-panel.kdl
2025-11-19 09:40:37.684  INFO Primary window confirmed on target screen screen=0
2025-11-19 09:40:37.702  INFO refreshing schematic path="drone.kdl"
2025-11-19 09:40:37.765  INFO Loaded secondary schematic path=motor-panel.kdl
2025-11-19 09:40:37.765  INFO Loaded secondary schematic path=rate-control-panel.kdl
2025-11-19 09:40:37.768  INFO Primary window confirmed on target screen screen=0
2025-11-19 09:40:37.804  INFO refreshing schematic path="drone.kdl"
2025-11-19 09:40:37.854  INFO Loaded secondary schematic path=motor-panel.kdl
2025-11-19 09:40:37.855  INFO Loaded secondary schematic path=rate-control-panel.kdl
2025-11-19 09:40:37.858  INFO Primary window confirmed on target screen screen=0
2025-11-19 09:40:37.905  INFO refreshing schematic path="drone.kdl"
2025-11-19 09:40:37.946  INFO Loaded secondary schematic path=motor-panel.kdl
2025-11-19 09:40:37.946  INFO Loaded secondary schematic path=rate-control-panel.kdl
2025-11-19 09:40:37.949  INFO Primary window confirmed on target screen screen=0
2025-11-19 09:40:38.006  INFO refreshing schematic path="drone.kdl"
2025-11-19 09:40:38.037  INFO Loaded secondary schematic path=motor-panel.kdl
2025-11-19 09:40:38.037  INFO Loaded secondary schematic path=rate-control-panel.kdl
2025-11-19 09:40:38.040  INFO Primary window confirmed on target screen screen=0
2025-11-19 09:40:38.108  INFO refreshing schematic path="drone.kdl"
2025-11-19 09:40:38.125  INFO Loaded secondary schematic path=motor-panel.kdl
2025-11-19 09:40:38.126  INFO Loaded secondary schematic path=rate-control-panel.kdl
2025-11-19 09:40:38.129  INFO Primary window confirmed on target screen screen=0
2025-11-19 09:40:38.209  INFO refreshing schematic path="drone.kdl"
2025-11-19 09:40:38.216  INFO Loaded secondary schematic path=motor-panel.kdl
2025-11-19 09:40:38.217  INFO Loaded secondary schematic path=rate-control-panel.kdl
2025-11-19 09:40:38.220  INFO Primary window confirmed on target screen screen=0
2025-11-19 09:40:38.310  INFO refreshing schematic path="drone.kdl"
2025-11-19 09:40:38.311  INFO Applying primary window rect rect=WindowRect { x: 22, y: 15, width: 57, height: 68 } width_px=2189 height_px=1469 x=845 y=324
```

zxq82lm:
> @zxq82lm thanks for the rundown. I'm doing some generic testing now before looking closer at the code. Overall the multi-window stuff is awesome. Great work!
> 
> I am seeing some interesting behavior. If do the following steps:
> 
> 1. Load drone example
> 2. Move windows
> 3. Save schematic
> 4. Close
> 5. Load drone example again
> 6. Save _again_
> 
> Elodin goes into some sort of a loop. I'm using i3wm, which is pretty nonstandard, so that might be the problem. I'm curious if you see the same behavior on Ubuntu?
> 
> Here's what I'm see in the logs during the loop:
> 
> ```
> 2025-11-19 09:40:37.269  INFO Applying primary window rect rect=WindowRect { x: 22, y: 15, width: 57, height: 68 } width_px=2189 height_px=1469 x=845 y=324
> 2025-11-19 09:40:37.270  INFO Confirmed screen assignment (sync) screen=0 path=motor-panel.kdl
> 2025-11-19 09:40:37.270  INFO Confirmed screen assignment (sync) screen=0 path=rate-control-panel.kdl
> 2025-11-19 09:40:37.273  INFO refreshing schematic path="drone.kdl"
> 2025-11-19 09:40:37.352  INFO Loaded secondary schematic path=motor-panel.kdl
> 2025-11-19 09:40:37.353  INFO Loaded secondary schematic path=rate-control-panel.kdl
> 2025-11-19 09:40:37.355  INFO Primary window confirmed on target screen screen=0
> 2025-11-19 09:40:37.374  INFO refreshing schematic path="drone.kdl"
> 2025-11-19 09:40:37.438  INFO Loaded secondary schematic path=motor-panel.kdl
> 2025-11-19 09:40:37.439  INFO Loaded secondary schematic path=rate-control-panel.kdl
> 2025-11-19 09:40:37.442  INFO Primary window confirmed on target screen screen=0
> 2025-11-19 09:40:37.476  INFO refreshing schematic path="drone.kdl"
> 2025-11-19 09:40:37.577  INFO refreshing schematic path="drone.kdl"
> 2025-11-19 09:40:37.603  INFO Loaded secondary schematic path=motor-panel.kdl
> 2025-11-19 09:40:37.604  INFO Loaded secondary schematic path=rate-control-panel.kdl
> 2025-11-19 09:40:37.615  INFO Primary window confirmed on target screen screen=0
> 2025-11-19 09:40:37.681  INFO Loaded secondary schematic path=motor-panel.kdl
> 2025-11-19 09:40:37.681  INFO Loaded secondary schematic path=rate-control-panel.kdl
> 2025-11-19 09:40:37.684  INFO Primary window confirmed on target screen screen=0
> 2025-11-19 09:40:37.702  INFO refreshing schematic path="drone.kdl"
> 2025-11-19 09:40:37.765  INFO Loaded secondary schematic path=motor-panel.kdl
> 2025-11-19 09:40:37.765  INFO Loaded secondary schematic path=rate-control-panel.kdl
> 2025-11-19 09:40:37.768  INFO Primary window confirmed on target screen screen=0
> 2025-11-19 09:40:37.804  INFO refreshing schematic path="drone.kdl"
> 2025-11-19 09:40:37.854  INFO Loaded secondary schematic path=motor-panel.kdl
> 2025-11-19 09:40:37.855  INFO Loaded secondary schematic path=rate-control-panel.kdl
> 2025-11-19 09:40:37.858  INFO Primary window confirmed on target screen screen=0
> 2025-11-19 09:40:37.905  INFO refreshing schematic path="drone.kdl"
> 2025-11-19 09:40:37.946  INFO Loaded secondary schematic path=motor-panel.kdl
> 2025-11-19 09:40:37.946  INFO Loaded secondary schematic path=rate-control-panel.kdl
> 2025-11-19 09:40:37.949  INFO Primary window confirmed on target screen screen=0
> 2025-11-19 09:40:38.006  INFO refreshing schematic path="drone.kdl"
> 2025-11-19 09:40:38.037  INFO Loaded secondary schematic path=motor-panel.kdl
> 2025-11-19 09:40:38.037  INFO Loaded secondary schematic path=rate-control-panel.kdl
> 2025-11-19 09:40:38.040  INFO Primary window confirmed on target screen screen=0
> 2025-11-19 09:40:38.108  INFO refreshing schematic path="drone.kdl"
> 2025-11-19 09:40:38.125  INFO Loaded secondary schematic path=motor-panel.kdl
> 2025-11-19 09:40:38.126  INFO Loaded secondary schematic path=rate-control-panel.kdl
> 2025-11-19 09:40:38.129  INFO Primary window confirmed on target screen screen=0
> 2025-11-19 09:40:38.209  INFO refreshing schematic path="drone.kdl"
> 2025-11-19 09:40:38.216  INFO Loaded secondary schematic path=motor-panel.kdl
> 2025-11-19 09:40:38.217  INFO Loaded secondary schematic path=rate-control-panel.kdl
> 2025-11-19 09:40:38.220  INFO Primary window confirmed on target screen screen=0
> 2025-11-19 09:40:38.310  INFO refreshing schematic path="drone.kdl"
> 2025-11-19 09:40:38.311  INFO Applying primary window rect rect=WindowRect { x: 22, y: 15, width: 57, height: 68 } width_px=2189 height_px=1469 x=845 y=324
> ```
----
@anderspitman Thanks a ton for catching this,it’s a real bug. 
I can reproduce it on Ubuntu (but not on macOS). 
Root cause: when you move windows and save, Elodin rewrites drone.kdl plus the two secondary .kdl files. We have an inotify watcher to auto-reload on changes; under i3/Ubuntu it emits several notifications in a row for the same file, and our ignore logic only blocks the first one. The remaining events trigger reloads, which rewrite the files again, and so on. I’ll ship a fix in this branch as soon as I’ve coded and tested it.

anderspitman:
Is it happening just with i3, or also the default Ubuntu window manager?

zxq82lm:
> Is it happening just with i3, or also the default Ubuntu window manager?

Probably reproduced with the default Ubuntu window manager (still investigating).

anderspitman:
@zxq82lm I see you pushed some more commits. Just let me know when this is ready for another look.

zxq82lm:
> @zxq82lm I see you pushed some more commits. Just let me know when this is ready for another look.

Sure, still analyzing the issue. I’ll let you know as soon as it’s ready.

libs/elodin-editor/src/ui/command_palette/palette_items.rs
```diff
@@ -859,18 +859,22 @@ pub fn save_schematic() -> PaletteItem {
         |_name: In<String>,
          db_config: Res<DbConfig>,
          schematic: Res<CurrentSchematic>,
-         secondary: Res<CurrentSecondarySchematics>| {
+         secondary: Res<CurrentSecondarySchematics>,
+         mut live_reload: ResMut<SchematicLiveReloadRx>| {
             match db_config.schematic_path() {
                 Some(path) => {
+                    live_reload.guard_for(Duration::from_millis(2000));
```
anderspitman: Is the live reload guard/ignore stuff still necessary, or can it go away similar to the window lock functionality? Would be really nice to avoid timeouts for things like this.

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -998,13 +2216,79 @@ fn set_secondary_camera_viewport(
...
+fn dedup_camera_orders(mut cameras: Query<(Entity, &mut Camera)>) {
+    // Collect active window-targeted cameras with their current order.
+    let mut entries: Vec<(Entity, String, isize)> = cameras
+        .iter()
+        .filter_map(|(entity, camera)| {
+            if !camera.is_active {
+                return None;
+            }
+            if let RenderTarget::Window(window_ref) = &camera.target {
+                return Some((entity, format!("{:?}", window_ref), camera.order));
```
anderspitman: Since we're using it for comparison, I think we might want to use [NormalizedWindowRef](https://docs.rs/bevy/latest/bevy/window/struct.NormalizedWindowRef.html) instead of format! here.

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -998,13 +2216,79 @@ fn set_secondary_camera_viewport(
...
+                    physical_position: UVec2::new(0, 0),
+                    physical_size: UVec2::new(1, 1),
+                    depth: 0.0..1.0,
+                });
+            }
+        }
+    }
+}
+
+fn dedup_camera_orders(mut cameras: Query<(Entity, &mut Camera)>) {
```
anderspitman: Why is this system necessary? I've been working on camera ordering with the bevy 0.17 update, and I'm worried this might interfere with our ability to set up the orders the way we want.

libs/elodin-editor/src/lib.rs
```diff
@@ -170,6 +180,17 @@ impl Plugin for EditorPlugin {
                     .disable::<LogPlugin>()
                     .build(),
             )
+            .add_plugins(LogPlugin {
+                // Reduce wgpu/present_frames noise on Linux to curb UI lag from log spam.
+                // Only set on Linux; other platforms keep defaults.
+                level: Level::INFO,
+                filter: if cfg!(target_os = "linux") {
```
anderspitman: Can this be a compile-time check instead of runtime?

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -7,38 +8,64 @@ use bevy::{
...
+pub(crate) const DEFAULT_SECONDARY_RECT: WindowRect = WindowRect {
+    x: 10,
+    y: 10,
+    width: 80,
+    height: 80,
+};
+const SCREEN_RELAYOUT_MAX_ATTEMPTS: u8 = 5;
+const SCREEN_RELAYOUT_TIMEOUT: Duration = Duration::from_millis(2000);
+const FULLSCREEN_EXIT_CONFIRMATION_TIMEOUT: Duration = Duration::from_millis(500);
+const LINUX_MULTI_WINDOW: bool = cfg!(target_os = "linux");
```
anderspitman: Same as above. Can we make this compile-time?

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -7,38 +8,64 @@ use bevy::{
...
+const SECONDARY_RECT_CAPTURE_LOAD_GUARD: Duration = Duration::from_millis(2500);
+const SECONDARY_RECT_CAPTURE_STABILIZE_GUARD: Duration = Duration::from_millis(400);
+const PRIMARY_VIEWPORT_ORDER_BASE: isize = 0;
+const PRIMARY_GRAPH_ORDER_BASE: isize = 100;
+// Secondary cameras are offset high to avoid clashing with primary/UI cameras.
+const SECONDARY_GRAPH_ORDER_BASE: isize = 1000;
+const SECONDARY_GRAPH_ORDER_STRIDE: isize = 50;
+const MACOS_PRIMARY_ORDER_OFFSET: isize = if cfg!(target_os = "macos") { 1000 } else { 0 };
```
anderspitman: Compile-time

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -528,28 +378,30 @@ impl RootWidgetSystem for MainLayout<'_, '_> {
     ) {
         let state_mut = state.get_mut(world);
 
-        let mut contexts = state_mut.contexts;
-        let images = state_mut.images;
+        let _contexts = state_mut.contexts;
```
anderspitman: These still necessary?

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -528,28 +378,30 @@ impl RootWidgetSystem for MainLayout<'_, '_> {
...
-            .frame(egui::Frame::NONE)
-            .show(ctx, |ui| {
-                ui.add_widget::<timeline::TimelinePanel>(world, "timeline_panel");
-                ui.add_widget_with::<tiles::TileSystem>(world, "tile_system", None);
-            });
+        #[cfg(target_os = "macos")]
+        let frame = {
+            let mut frame = egui::Frame::new();
+            // Leave a small top inset so native titlebar controls don't overlap egui.
+            frame.inner_margin.top = 32;
```
anderspitman: Is 32 going to look good on all platforms?

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -663,44 +533,1347 @@ fn sync_secondary_windows(
...
+            if let Ok(mut camera) = cameras.get_mut(graph) {
+                camera.is_active = false;
             }
         }
 
         let title = compute_secondary_window_title(state);
 
+        // Try to pre-size and pre-position the window to its target rect to avoid an
+        // extra resize pass (and the resulting swapchain churn).
+        let (resolution, position, _pre_applied_rect, pre_applied_screen) = if let Some(rect) =
```
anderspitman: Can `_pre_applied_rect` be removed?

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -663,44 +533,1347 @@ fn sync_secondary_windows(
...
+            },
+            ..Default::default()
+        };
+
         let window_entity = commands
-            .spawn((
-                Window {
-                    title,
-                    resolution: WindowResolution::new(640.0, 480.0),
-                    present_mode: PresentMode::AutoVsync,
```
anderspitman, line 892: Do we not want Vsync and buttons anymore?

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -663,44 +533,1347 @@ fn sync_secondary_windows(
...
+        size_w = width_px,
+        size_h = height_px,
+        pos_x = x,
+        pos_y = y,
+        "Applied rect"
+    );
+    true
+}
+
+fn assign_window_to_screen(
```
anderspitman, line 1285: Some of your functions use "screen" and some use "monitor", I believe to refer to the same thing (please correct me if I'm wrong about that). We should probably be consistent.

libs/elodin-editor/src/ui/schematic/load.rs
```diff
@@ -595,14 +629,159 @@ pub fn graph_label(graph: &Graph) -> String {
...
+pub struct SchematicLiveReloadRx {
+    receiver: Option<flume::Receiver<SchematicReloadEvent>>,
+    ignored_paths: HashMap<PathBuf, usize>,
+    #[cfg(target_os = "linux")]
+    load_guard_until: Option<Instant>,
+    #[cfg(target_os = "linux")]
+    last_loaded: HashMap<PathBuf, SystemTime>,
+}
+
+impl SchematicLiveReloadRx {
```
anderspitman, line 713: See above. Do we still need all this or can we simplify our data flow/source of truth logic to avoid needing it?

anderspitman (dismissed):
@zxq82lm only had time for a few flyby comments until next week, but hopefully at least a little useful.

libs/elodin-editor/src/ui/mod.rs
zxq82lm: Good catch! We’ve now standardized it in the code: “monitor” refers to the monitoring/supervision object in Elodin, and “screen” refers to a physical screen. Thank you.

libs/elodin-editor/src/ui/mod.rs
zxq82lm: We now force Fifo (guaranteed vsync) on all platforms and re-enable native close/min/max buttons on secondary windows.

libs/elodin-editor/src/ui/mod.rs
zxq82lm: Thanks for the pointer -> I didn’t know about NormalizedWindowRef. 
Switched to it so window target comparisons are stable without relying on strings.

libs/elodin-editor/src/lib.rs
zxq82lm: Thank you. 
We unified the log level across platforms and now filter wgpu_core/wgpu_hal to warn—no more OS-specific log handling.

libs/elodin-editor/src/ui/mod.rs
zxq82lm: Converted the platform flags to compile-time cfg (Linux vs non-Linux).

libs/elodin-editor/src/ui/mod.rs
zxq82lm: Converted the platform flags to compile-time cfg (Linux vs non-Linux).

libs/elodin-editor/src/ui/mod.rs
zxq82lm: The unused pre-applied rect binding is removed.

libs/elodin-editor/src/ui/mod.rs
zxq82lm: The unused frame inset is removed.

libs/elodin-editor/src/ui/mod.rs
zxq82lm: This is no longer necessary; I’ve dropped it.

libs/elodin-editor/src/ui/mod.rs
zxq82lm: We removed dedup_camera_orders. It was a guard to avoid flicker from order collisions (mostly on the gizmo). The root cause was how the gizmo was assigned on secondary viewports; we fixed it by giving the gizmo dedicated orders and a stable per-window viewport (primary/secondary), so the dedup system isn’t needed anymore.

libs/elodin-editor/src/ui/command_palette/palette_items.rs
zxq82lm: When we save, we write both the main KDL and the secondary window KDLs. On Linux, notify emits multiple events during those writes. Without the guard, live reload can trigger mid-write and reload a half-written file, which can lead to inconsistent schematics, UI flicker, and even “re-creating” secondary windows while they’re still being saved. The guard just suppresses notifications long enough for the writes to finish so we don’t reload prematurely.

I’m not entirely convinced this is the best approach; there might be a cleaner one:
- Instead of a timeout, we could use a cross-thread “saving” flag. The reload thread would check a shared Arc<AtomicBool>/mutex before handling each event, and every save path (main + secondary) would use an RAII guard to set/clear it reliably (even on errors). We’d also have to decide whether to drop or buffer events that arrive while the flag is set. That’s all doable, but it adds coordination and more edge cases. Functionally, the current implementation has been stable and satisfactory in my testing so far.

What’s your opinion on these two approaches? @anderspitman @shanecelis @x46085

x46085: I’m good with it. I don’t see the harm in this case, and to solve any concern over brittleness, just make it like 5 seconds. Unlikely to ever be violated

anderspitman: @x46085 so that would mean changes to the KDL files are ignored for 5 seconds? Are we sure a user wouldn't want to have a workflow where they're quickly switching back and forth between WYSIWYG changes and editing the KDL directly?

@zxq82lm so the problem is that we need to process the updates to all the files as one atomic group, but the events are coming in separately? Assuming I'm understanding that correctly, I think we might be working around a deeper issue, that if solved may allow us to simplify the approach.

I think the root problem is that saving the schematic from the GUI editor triggers multiple file change notifications that we need to handle. But those notifications are superfluous. The changes originated from GUI editor, so there should be no need to process them. The KDL is the source of truth, but if the editor already matches the source of truth, no changes are necessary. Could we just keep the serialized string of each KDL file (or even a hash) in memory, and whenever a notification for that KDL path comes in, we read the file contents, checked if it's changed, and if not ignore the notification?

This is a bit off the cuff so please let me know if I'm missing something.

libs/elodin-editor/src/ui/mod.rs
anderspitman: Is there a reason we went with FIFO over AutoVsync?

zxq82lm: > Is there a reason we went with FIFO over AutoVsync?

FIFO guarantees a vsync’d swapchain everywhere (required to be supported), so no tearing and consistent timing across platforms. AutoVsync can fall back to immediate/mailbox if the backend or compositor decides, which can reintroduce tearing or varying latency. We picked FIFO to lock in “always vsync, predictable” behavior.

libs/elodin-editor/src/ui/command_palette/palette_items.rs
zxq82lm: > @zxq82lm so the problem is that we need to process the updates to all the files as one atomic group, but the events are coming in separately? Assuming I'm understanding that correctly, I think we might be working around a deeper issue, that if solved may allow us to simplify the approach.
> 
> I think the root problem is that saving the schematic from the GUI editor triggers multiple file change notifications that we need to handle. But those notifications are superfluous. The changes originated from GUI editor, so there should be no need to process them. The KDL is the source of truth, but if the editor already matches the source of truth, no changes are necessary. Could we just keep the serialized string of each KDL file (or even a hash) in memory, and whenever a notification for that KDL path comes in, we read the file contents, checked if it's changed, and if not ignore the notification?
> 
> This is a bit off the cuff so please let me know if I'm missing something.

- Your read is right: on Linux (only) we write a bundle (main + secondary KDLs) and events arrive separately, so the guard prevents reloading half-written files during that burst. Hash/serialized comparison could cut redundant reloads, but you still must avoid reading/hashing mid-write; a mixed approach is possible (brief silence, then hash-filter GUI saves). 
- Limited gain: the guard already covers the risky cases and stays simple. Hash/cache adds I/O/sync complexity and doesn’t remove the need to protect the write phase, so if things are stable the guard is a reasonable compromise. Note: this guard logic is Linux-only and narrowly scoped to our own saves.

libs/elodin-editor/src/ui/schematic/load.rs
zxq82lm: It’s the same concern as the guard above (only on Linux): last_loaded is used to filter bursty notify events alongside the guard.

libs/elodin-editor/src/ui/command_palette/palette_items.rs
anderspitman: > but you still must avoid reading/hashing mid-write

Ah so you're saying that even for a single write, multiple notifications are being emitted, probably due to multiple syscalls under the hood?

zxq82lm: Yes. On Linux a single save can fire a burst of _inotify_ events (write → close_write, plus rename/attrib if the editor saves via a temp file). Since we write the main and secondary KDLs separately, you also get distinct events for each. So even for one save you can see multiple signals; you still need to guard the write window, a hash alone isn’t enough.

anderspitman: My proposal should fix the multi-file issue. Say we save and it writes to 3 files. We'll receive notifications for each of them separately, but in each case we would see that the contents of the file matches what we already have (because we are the ones who changed them), so no further processing happens.

The partial writes is trickier...

zxq82lm: > My proposal should fix the multi-file issue. Say we save and it writes to 3 files. We'll receive notifications for each of them separately, but in each case we would see that the contents of the file matches what we already have (because we are the ones who changed them), so no further processing happens.
> 
> The partial writes is trickier...

I suggest opening a follow-up issue post-merge including:
- Explore your idea: cache hash/serialized KDL per file and, on notify, read/compare to skip GUI-originated saves.
- Keep a brief protection against partial writes (short guard or “saving” flag) so we never read/hash a half-written file.

That would let us simplify the current guard if it pans out? @anderspitman

anderspitman: https://github.com/elodin-sys/elodin/issues/325

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -217,16 +271,64 @@ impl Plugin for UiPlugin {
...
+                track_secondary_window_geometry.after(confirm_secondary_screen_assignment),
+            )
+            .add_systems(
+                Update,
+                capture_secondary_window_screens.after(track_secondary_window_geometry),
+            )
+            .add_systems(
+                Update,
+                capture_primary_window_layout.after(capture_secondary_window_screens),
+            )
```
shanecelis: These system orderings are pretty tricky to follow. Is there a reason to not do `(a, b, c).chain()` instead?

shanecelis (changes requested):
This works, but I'm a little concerned with some of the changes. I have added two commits (with permission, of course). 

One of the pain points I saw with this code is that there's a struggle to express essentially an asynchronous operation within a synchronous code block. I see this in `apply_secondary_window_screens` for instance. I have started a refactor to use `bevy_defer`, which provides an asynchronous API, which I think will help this block of code a lot.

I have not gone through and removed the fields from state that are no longer necessary. It compiles but there is a `todo!` that remains to be resolved.

Finally, there is a fair amount of duplication between the Primary and Secondary windows implementation. For instance there is another system `apply_primary_window_layout` that has the same kind of issues addressed in the refactor commit I mentioned above. It would be lovely to try and reduce some of this code and struct duplication. I wonder if an enum that was `enum EWindow { Primary { ... }, Secondary { ... } }` might help. We had two enums that had the same variants but one was named `SecondaryWindowRelayoutPhase`. I'm happy to discuss it further on discord as you prefer.

I'd encourage you to try leveraging `bevy_defer`. It's a very useful library.

shanecelis:
After directly adding my commits to this branch, I realized I would have preferred to add them to a separate PR so the changes are easier to inspect. Then I realized I was just a push and force-push away from having exactly that. So my big refactor of this PR is in PR #331.

x46085 (approved):
Make it so!

libs/elodin-editor/src/ui/command_palette/palette_items.rs
```diff
@@ -856,30 +834,41 @@ pub fn save_schematic() -> PaletteItem {
...
-                    PaletteEvent::Exit
-                }
-                None => PalettePage::new(vec![save_schematic_inner()]).into(),
-            }
+        |_name: In<String>, mut commands: Commands| {
+            // Refresh primary path + descriptors, rebuild schematics, then serialize in a dedicated system.
+            commands.run_system_cached(crate::ui::update_primary_descriptor_path);
+            commands.run_system_cached(crate::ui::capture_window_screens_oneoff);
+            commands.run_system_cached(crate::ui::schematic::tiles_to_schematic);
+            commands.run_system_cached(save_schematic_now);
```
shanecelis, line 842: Very cool to see the `run_system` in use!

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -826,130 +640,950 @@ fn readable_label_color(color: Color) -> Color32 {
...
+        }
+
+        AsyncWorld.yield_now().await;
+    }
+    Ok(false)
+}
+
+#[cfg(not(target_os = "macos"))]
+#[cfg(not(target_os = "macos"))]
+#[cfg(not(target_os = "macos"))]
```
shanecelis, line 948: Cfg triple not necessary, right?

shanecelis (approved):
Working well on macOS!

Issue #266, opened 2025-10-27
Rocket flickers at beginning


shanecelis:
There is a flicker in the rocket example as the video shows.

https://github.com/user-attachments/assets/e30bed8d-5e83-4c18-9e0b-23ec83652ad2

It's only present in the rocket example that I've seen. If one pauses and steps through time, it does not show up. Once the simulation has run, one can scrub over time and it does not appear. My hunch would be that there is some position or orientation synchronization that I missed when I fixed the vector arrows distortion #217.

shanecelis:
This bug is present in just about every sim. There is a flicker on moving objects and moving cameras and still cameras as far as I can tell. I've recorded my screen to capture it. I think I've eliminated big_space as a problem. When I exclude the bevy_editor_cam plugin, the flicker goes away, so I will look into how that crate is scheduling things.

shanecelis:
Finding the one frame where this flicker happens is troublesome. Rather than post a video here and letting you watch intently, I've created a GIF from the two frames. Click play to see the two frames.

![Image](https://github.com/user-attachments/assets/8458e905-1445-40e2-bbe2-a13d568e0455)

shanecelis:
A fix is to change `bevy_editor_cam` from using the `PreUpdate` schedule to `Update` in each of its plugins. Here is a video of it in action.

https://github.com/user-attachments/assets/d3a91753-a2a1-4432-8245-0033a5bec583

shanecelis:
Posted [issue](https://github.com/aevyrie/bevy_editor_cam/issues/51) to bevy_editor_cam to solicit maintainer feedback.

shanecelis:
bevy_editor_cam doesn't uniformly use `PreUpdate` throughout its plugins. I found a happenstance fix when I moved adding the bevy_editor_cam plugin below where we add our systems: The flicker went away. That one-line change does not inspire confidence since Bevy's scheduler is free to order them differently in another release, on another platform, on a different day, etc. But it did inspire me to find the conflicting system and label it in a system set. Then I added a constraint that we run our camera positioning system before bevy_editor_cam's, which fixes the issue.

shanecelis:
Fixed in 0.15.4.

PR #303, opened 2025-11-07
Run app using `cargo` from top-level with way less verbosity

shanecelis:
# Problem

Running the editor in the dev environment can be quite verbose.

```sh
cargo run --features elodin-editor/inspector --manifest-path=apps/elodin/Cargo.toml editor examples/rocket/main.py
```

## Dismissed Solution

Make "apps/elodin" the default workspace so that a simple `cargo run` will run the editor. 

```sh
cargo run -- editor examples/rocket/main.py
```

The problem is this would make `cargo test` only run apps/elodin's tests. To run all tests one would have to do `cargo test --workspace`, which is not a rake we need to step on.

## Solution

Use the "run --bin" or an alias, both are available for "elodin" and "elodin-db".

### `run --bin`
```sh
cargo run --bin elodin editor examples/rocket/main.py
```

### Alias
```sh
cargo elodin editor examples/rocket/main.py
```

This PR adds these aliases and updates the READMEs to use them.

# Problem 2

The feature flags are hard to remember. I introduced "elodin-editor/inspector" a ways back, but it'd be nice to lose the somewhat meaningless prefix of "elodin-editor".

# Solution 

Add "inspector" to apps/elodin so the command can be the following:

```sh
cargo elodin --features inspector editor examples/rocket/main.py

```

This PR also adds a "debug" flag that enables the big_space debug floating origin visualization to try and track down this issue #266.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,12 +78,11 @@ Open the Elodin editor in a new nix develop shell and connect to the local serve
 # In a new terminal
 nix develop
 # build the SDK python wheel
-cd libs/nox-py && \
 uv venv --python 3.12 && \
 source .venv/bin/activate && \
-uvx maturin develop --uv
+uvx maturin develop --uv --manifest-path=libs/nox-py/Cargo.toml
 # use the newly built wheel
-python3 ../../examples/rocket/main.py run
+python3 examples/rocket/main.py run
 ```
 
 Open the Elodin editor and connect to the local server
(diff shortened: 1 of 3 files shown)


apps/elodin/README.md
```diff
@@ -7,29 +7,26 @@ Install the editor from the [releases](https://github.com/elodin-sys/elodin/rele
...
-
 # Create a new virtual environment using uv
 uv venv
 
 # Activate the environment (Bash)
 source .venv/bin/activate
 
 # Build and install nox-py in development mode
-uvx maturin develop --uv
+uvx maturin develop --uv --manifest-path=libs/nox-py/Cargo.toml
```
x46085, line 17: You should update the `nix develop` shell help text to match (/nix/shell.nix)

shanecelis: Good catch. I found some other strays too in the docs.

Issue #281, opened 2025-11-04
Clicking on the axes widget sometimes causes the camera to start spinning endlessly


anderspitman:
I've been able to get it into this state twice, once by accident and once intentionally. Haven't spent more time on it beyond that.

shanecelis:
Good find. Haven't seen this yet. The widget needs work anyhow #273.

PR #315, opened 2025-11-14
fix: The axes widget no longer spins indefinitely

zxq82lm:
## PR objective
The purpose of this PR is to address this issue [#281](https://github.com/elodin-sys/elodin/issues/281).

## How to test
- In the drone scene, click/drag/release the *axes widget* points several times (fast motions, switching axes, releasing outside the widget).
- After each drag the camera must stop immediately: no orbit should continue on its own.
- The fix confines orbit to a single drag session and calls `end_move` when the drag ends; stress the *axes widget* the way it used to break and confirm no infinite rotation reappears.

### DIFF
--- a/libs/elodin-editor/src/plugins/camera_anchor.rs
+++ b/libs/elodin-editor/src/plugins/camera_anchor.rs
@@ -7,12 +7,7 @@ use bevy::transform::components::Transform;
 /// because it contains invalid scale values), which signals the caller to let
 /// `EditorCam` reuse its last known depth instead of corrupting it with NaNs.
 pub fn camera_anchor_from_transform(transform: &Transform) -> Option<DVec3> {
-    let anchor = transform
-        .compute_matrix()
-        .as_dmat4()
-        .inverse()
-        .w_axis
-        .xyz();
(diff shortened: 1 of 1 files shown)


libs/elodin-editor/src/plugins/camera_anchor.rs
```diff
@@ -0,0 +1,21 @@
...
+///
+/// This falls back to `None` if the transform cannot be inverted (for example,
+/// because it contains invalid scale values), which signals the caller to let
+/// `EditorCam` reuse its last known depth instead of corrupting it with NaNs.
+pub fn camera_anchor_from_transform(transform: &Transform) -> Option<DVec3> {
+    let anchor = transform
+        .compute_matrix()
+        .as_dmat4()
+        .inverse()
+        .transform_point3(DVec3::ZERO);
```
shanecelis: For this `transform_point3`, you can read off the fourth column from the matrix. I double-checked against the implementation of [`transform_point3`](https://docs.rs/glam/0.30.8/src/glam/f64/dmat4.rs.html#1169), and it had no surprises. So instead of `transform_point`, you can put `w_axis.xyz()`.

I also confirmed via Mathematica.

Screenshot 2025-11-24 at 2 48 45 PM

PR #316, opened 2025-11-17
fix: Ctrl+C able to interrupt simulation on Linux

zxq82lm:
## PR objective
- Address this issue [#312](https://github.com/elodin-sys/elodin/issues/312)
- Reproduced with:
```
cargo run --manifest-path=apps/elodin/Cargo.toml editor examples/drone/main.py
```

## Branch description
Two small changes: 
- Add a tiny timeout (50 ms) in the `io_uring` loop so it never stays stuck
- Ignore EINTR/ETIME and exit with code 130 on SIGINT. 

Effect: on Linux the loop wakes up regularly, sees Ctrl+C, and stops the simulation.

PR #321, opened 2025-11-21
Refactor Query Panel, support derived component data viz

x46085:
Use Query Panel to show derived data, such as "angle of attack":

https://github.com/user-attachments/assets/b713db23-7863-4bea-9877-76851786ef39

### New Features
- **Literal arithmetic in EQL**: Support for operations between expressions and literals (e.g., `formula() - 90.0`)
  - Modified `to_table()`, `to_sql()`, `to_select_part()` to handle `BinaryOp` with literal operands
  - Added integration tests for literal arithmetic patterns

### Formula Testing
- Added comprehensive edge case tests for: `abs`, `arccos`, `atan2`, `clip`, `degrees`, `sign`, `sqrt`
- Verified formula chaining, negative values, and boundary conditions
- Fixed `atan2` argument order handling and warnings

### Query Plot Fixes
- **Disabled Y-axis auto-normalization** - preserves intentional offsets in queries (was removing `- 90.0` effect)
- **Initialization artifact detection** - skips duplicate-timestamp initial points that distort graph scaling
- Fixed clippy warning in timestamp comparison loop

### Validation
- Verified rocket angle-of-attack calculation now matches Python implementation exactly

## Changes Made

- **Fixed `sign()` EQL formula**: Replaced PostgreSQL `sign()` with a DataFusion-compatible SQL `CASE` statement (`CASE WHEN x > 0 THEN 1 WHEN x < 0 THEN -1 ELSE 0 END`) to resolve "Invalid function 'sign'" errors.
- **Fixed `arccos()` EQL formula**: Added automatic input clipping to [-1, 1] using `GREATEST(-1.0, LEAST(1.0, ...))` to match Python's domain requirements and prevent NaN/errors from out-of-range values.
- **Updated Angle of Attack EQL formula**: Matches Python calculation `arccos(dot(u, [-1,0,0]) / norm(u)) * -sign(w)` using `arccos()` and `sign()` formulas.
- **Added tests**: Tests for `arccos()` clipping behavior, chained `arccos().degrees()` patterns, and division operations.

## Problems Solved

- Resolved DataFusion planning error for `sign()` function
- Fixed 90-degree offset in Angle of Attack EQL calculation (now matches component calculation)
- Ensured EQL `arccos()` matches Python's `arccos()` behavior for same inputs
- Added safeguards against NaN/invalid values from out-of-range `arccos()` inputs

---

> [!NOTE]
> Adds new EQL formulas and literal arithmetic, refactors Query Plot to render EQL-derived values with relative time, and updates the rocket example to compute/show angle of attack from `v_body`.
> 
> - **DB/EQL**:
>   - Add formulas: `atan2`, `degrees`, `clip`, `sqrt`, `abs`, `arccos` (with input clipping), `sign` (CASE-based).
>   - Support literal arithmetic in SQL generation; improve table/tuple resolution and expose `to_sql_time_field`.
>   - Expand README with grammar/examples; add extensive unit/integration tests.
> - **Editor (Query Plot & Graphs)**:
>   - Refactor to unified `TimeseriesPlot` with `PlotDataSource` (supports XY relative-time plots).
>   - Render EQL query results with time column prepended; proper scrub cursor, modal readouts, and current-time tick.
>   - Disable Y auto-normalization for queries; skip duplicate-timestamp init points; improved bounds/axis labeling.
>   - Wire scrub icon through tiles; immediate first refresh; component extraction from complex EQL.
> - **Rocket Example**:
>   - Add `v_body` component/system; add EQL `query_plot` panels for angle of attack and velocity magnitude.
>   - Update README with EQL usage, features, and controls; set `max_ticks=5000`.
> - **Infra**:
>   - Update `flake.lock`/`flake.nix`; simplify `shell.nix` (remove sccache).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c0e69d79a854451faa967772d74c35705aab7903. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/libs/elodin-editor/src/ui/plot/data.rs
+++ b/libs/elodin-editor/src/ui/plot/data.rs
@@ -570,7 +570,7 @@ impl XYLine {
             let chunk = gpu.as_index_chunk::<f32>(buf.cpu().len());
             for (i, index) in chunk.into_index_iter().enumerate() {
                 let absolute = global_index + i;
-                if absolute % step == 0 || absolute + 1 == total_points {
+                if absolute.is_multiple_of(step) || absolute + 1 == total_points {
                     view = append_u32(view, index);
                     count += 1;
                 }


x46085:
There is still a bug in here, where the plot line stops drawing after a while. Could use some help looking into that one.

zxq82lm:
> There is still a bug in here, where the plot line stops drawing after a while. Could use some help looking into that one.

- I can reproduce the issue (review).
- I might have found a possible explanation, but I’m not completely sure yet. INDEX_BUFFER_LEN is currently limited to 4096 in libs/elodin-editor/src/ui/plot/gpu.rs (line 61). It looks like Angle-Of-Attack and Velocity-Magnitude eventually hit that 4096-index limit: at that point append_u32 stops writing (the index buffer is full), but count keeps increasing. As a result, the GPU keeps reusing the same 4096 indices while we ask it to draw more, so the line appears to stop evolving / looks frozen.
---
@x46085 
On this branch https://github.com/elodin-sys/elodin/tree/pr-321-xy-decimation-fix, which is based on feature/adv-eql, there’s a proposed fix that I just tested ok.
image

- It introduces decimation (a technique I wasn’t familiar with before), i.e. distributing/sampling the indices so that we stay within the fixed GPU buffer size instead of pushing every single point.
The changes are in data.rs and gpu.rs.

x46085:
> * It introduces decimation (a technique I wasn’t familiar with before), i.e. distributing/sampling the indices so that we stay within the fixed GPU buffer size instead of pushing every single point.
>   The changes are in data.rs and gpu.rs.

All for it! I think it's possible that previously the idea was there would be multiple queries over time and the values would be appended each time. But I think this works for now.

Feel free to commit to my branch.

x46085:
@zxq82lm could you also just double check I didn't break the EQL table widget? I shouldn't have, but wanted to be sure.

zxq82lm:
> @zxq82lm could you also just double check I didn't break the EQL table widget? I shouldn't have, but wanted to be sure.
---
Just tested, didn’t see any issues: the widget opens correctly and the EQL/SQL queries execute fine.
image

x46085:
Beautiful. Feel free to approve and we’ll make it so!

zxq82lm (approved):
The features and new EQL formulas have been successfully tested.

---
> Side note: I’ve had some issues with Ctrl+C on macOS with some examples (rocket and ball), but that seems unrelated to this branch.

x46085:
> > Side note: I’ve had some issues with Ctrl+C on macOS with some examples (rocket and ball), but that seems unrelated to this branch.

Yes I'm seeing it now too, and it's very new to me. I don't see the cause, but I would guess this change introduced this issue on the Mac: https://github.com/elodin-sys/elodin/pull/316

PR #323, opened 2025-11-21
Support SITL Python module

shanecelis:
This PR is directed towards an integration effort to run Software In The Loop (SITL) testing in a Continuous Integration (CI) mode. I will start with the essential changes and then make note of incidental changes to support this effort.

## Support new `world.run(...)` arguments:

* `is_canceled` accepts a Python function that may return a boolean. If it
  returns `True` then the simulation is stopped. This functionality has been
  repurposed for SITL because it is called exactly once per step, so we use it
  to run the SITL after the sim writes to the database.
* `db_path` accepts a path to store a database. This will open a database if it
  exists or create one if it doesn't. Recommended usage is to provide a new
  database path each time.
* `interactive` flag accepts a boolean, true by default. It will keep the
  database up for interactive inspection via a database connection usually from
  the editor. When set to false, it will exit as soon as the simulation has
  reached its max tick.

## Support arguments to Lua scripts

One can provide a Lua script to `elodin-db` like this:

```sh
elodin-db lua print-results.lua
```

This PR adds the ability for the 'lua' subcommand to accept CLI arguments.

```sh
elodin-db lua print-results.lua -- 127.0.0.1:2240 rocket.commanded_deflect
```

```sh
$ cat print-results.lua
-- print_results.lua
--
-- Print the table of a given component in an Elodin database.
if arg[1] == "-h" or arg[1] == "--help" then
  print("Usage: elodin_db lua print_results.lua [HOST] [COMPONENT]")
  print("Print the table of a given component in an Elodin database.")
  return 1
end
host = arg[1] or "127.0.0.1:2240"
component = arg[2] or "rocket.commanded_deflect"
client = connect(host)
id = ComponentId(component)
client:get_time_series(id)
```

## Enhance error reporting

I added more information to `TimeRangeOutOfBounds` to ease debugging.

## Avoid alignment errors

The `DB` struct is aligned to 128 to support a caching crate. Python didn't like this, so I ended up boxing it in `Exec` to avoid that issue.

## Interrupt Signal Response

There is a lot of code that calls Python's `check_signals()` function. It seems there have been many requests to support canceling a process, yet our app can still become un-killable via Ctrl-C. Here's a note from that particular commit:

> We run our JAX World in a thread not limited by the Python Global
Interpreter Lock (GIL). However, prior to this commit, we acquire the
GIL each time we check if we should cancel by checking for Unix signals
using `py.check_signals()`. This commit changes our signal checking to
use a Rust library and sets an atomic bool flag that we then check for,
so we do not not need to acquire the GIL unless we're actually running a
Python function.

### DIFF
--- a/libs/nox-py/src/world_builder.rs
+++ b/libs/nox-py/src/world_builder.rs
@@ -288,14 +288,14 @@ impl WorldBuilder {
                             // Here we start the DB with an address.
                             elodin_db::Server::new(db_path, addr).unwrap(),
                             exec,
-                        ).run_with_cancellation(
+                        )
+                        .run_with_cancellation(
                             move || {
                                 if let Some(ref func) = is_canceled {
                                     Python::with_gil(|py| {
-                                        func.call0(py)
-                                            .and_then(|result| result.extract::<bool>(py))
+                                        func.call0(py).and_then(|result| result.extract::<bool>(py))
                                     })
-                                        .unwrap_or_else(|_| terminate_flag.load(Ordering::Relaxed))
+                                    .unwrap_or_else(|_| terminate_flag.load(Ordering::Relaxed))
                                 } else {
                                     terminate_flag.load(Ordering::Relaxed)
                                 }
@@ -303,7 +303,8 @@ impl WorldBuilder {
                             move |tick_count| {
                                 if let Some(ref func) = post_step {
                                     Python::with_gil(|py| {
-                                        let tick_count_py = tick_count.into_bound_py_any(py)
+                                        let tick_count_py = tick_count
+                                            .into_bound_py_any(py)
                                             .unwrap_or_else(|_| py.None().into_bound(py));
                                         if let Err(e) = func.call1(py, (tick_count_py,)) {
                                             tracing::warn!("post_step error {e}");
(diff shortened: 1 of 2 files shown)


libs/nox-py/src/world_builder.rs
```diff
@@ -257,16 +275,34 @@ impl WorldBuilder {
...
+                                            func.call0(py)
+                                                .and_then(|result| result.extract::<bool>(py))
+                                        })
+                                        .unwrap_or_else(|_| terminate_flag.load(Ordering::Relaxed))
+                                    } else {
+                                        terminate_flag.load(Ordering::Relaxed)
+                                    }
+                                }
+                            },
+                            interactive,
```
x46085, line 315: I think this change is the one that gives me the most pause. It's typically a code smell to overload functionality onto something that is unrelated to your intended affect. I don't see a clean and easy path to doing so however, so make it so.

shanecelis:
Yeah, that's fair. How about a `post_step` argument that accepts a callable?

x46085:
> Yeah, that's fair. How about a `post_step` argument that accepts a callable?

Yea anything that would be clearly purpose-built for it's intended use works for me.

PR #329, opened 2025-11-26
fix: handle Ctrl-C on macOS and Linux

zxq82lm:
## Context 
In the current main branch, Ctrl-C no longer works on macOS and the editor and/or simulation keeps running:
```
^C
2025-11-26 12:19:25.898  INFO Received Ctrl-C, shutting down
2025-11-26 12:19:25.898  INFO Waiting for sim process to exit
2025-11-26 12:19:25.972  WARN command receive error: receiving on a closed channel
```
The goal here is to get Ctrl-C working on both macOS and Linux.

## Tests
For spotting a leftover sim process:

Linux & macOS: 
```bash
ps -axo pid,ppid,state,command | grep -E 'main.py|s10' | grep -v grep
```

### DIFF
--- a/apps/elodin/src/cli/editor.rs
+++ b/apps/elodin/src/cli/editor.rs
@@ -88,7 +88,7 @@ impl Cli {
                             Ok(()) => {
                                 info!("Received Ctrl-C, shutting down");
                                 cancel_token.cancel();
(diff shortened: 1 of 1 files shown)


zxq82lm:
@codex analyze this PR.

PR #331, opened 2025-11-30
Multi window refactor

shanecelis:
I did some significant refactoring on PR #302. The code worked and this is a significant feature so kudos to Benoit for that, but there were some architectural issues I wanted to address. Because this touches on many Bevy and ECS topics, I'll outline what I did and why so it's clear to everyone on the team.

1. Removed `WindowManager`.

This manager was keyed on `WindowId(usize)` and kept the `WindowState` for the secondary windows. This is a common object-oriented design, but it's going against the grain in an Entity Component System (ECS). Instead we can...

2. Make `WindowId` and `WindowState` components.

Our key is `Entity`, and we keep the `WindowId` and the `WindowState` directly attached to our entity with the `Window` component rather than having a manager object that tries to keep the two in-sync.

I've made the main window use `WindowId(0)` and all other window IDs are auto-incremented from 1.

3. Use `bevy_defer` for async.

A lot of state was kept in structs to essentially handle asynchronous, multi-frame operations. Many a bespoke state machine was made. If you are bit by a case like this in the future, and don't have a suitable async solution like `bevy_defer`, I'd recommend doing something like this:

```rust
enum StateMachine {
     A,
     B { a: u32 },
     C { a: u32, b: String },
     D { a: String },
     E,
}

fn poor_mans_async(state: &mut StateMachine) {
    match state {
        StateMachine::A => {
            *state = StateMachine::B { a: 0 };
        }
        StateMachine::B { a } => {
            *a += 1;
            if a >= 100 {
                *state = StateMachine::C { a, b: String::from("hi") };
            }
        }
        // ...
    }
}

fn run_poor_mans_async() {
    let mut state = StateMachine::A;
    while state != StateMachine::E {
        poor_mans_async(&mut state);
    }
}
```

The above is essentially what async writes for us when we write code that looks like the following:

```rust
async fn rich_mans_async() -> Result<(), AccessWorld> {
     let mut a = 0;
     AsyncWorld.yield_now().await?;
     while a < 100 {
         a += 1;
         AsyncWorld.yield_now().await?;
     }
     let b = String::from("hi");
     // ...
}
```

Luckily, `bevy_defer` is an excellent library that allows us to access Bevy resources within an async context. It will not let you keep resources or references once you return from an async context via `.await`, so many times it'll hand you a Bevy resource to a closure to ensure that no references to it are kept between `.await`s.

Use `bevy_defer` if you have an operation that runs asynchronously over multiple frames that requires timing or coordination.

4. Prefer one-off systems to ignore flags.

There was a system that had many checks to ignore or block its evaluation. This makes sense because if you add it to Bevy's schedule, it will run every frame. In cases like that I encourage you to consider whether your system ought to be in the schedule at all. You can run Bevy systems as one-offs, or unscheduled systems. 

```rust
commands.run_system_cached(capture_window_screens_oneoff);
```

I added the "oneoff" suffix to indicate this system isn't part of the schedule, which may be a convention we want to adopt.

5. Re-evaluate type names that contain "Secondary" or "Primary".

We had two distinct code paths: one for the primary window, and another for secondary windows. This was codified in the type names that would sometimes impede them from code reuse. I have tried to unify these things where appropriate. There is a `window_id.is_primary()` boolean flag, and because the main window now uses `WindowId(0)` and has a `WindowState` it can be handled like any other window in many cases.

I think `secondary` and `primary` are fine field and variable names. 

6. Avoid allocations in systems when you can.

Because systems often run at 60 Hz, allocations should be avoided if possible. The following system allocates a new vector every frame.

```rust
fn system_a(windows: Query<Entity, With<&Window>>) {
    let w: Vec<Entity> = windows.iter().collect();

}
```
Here is a way that you can allocate one vector for the life of the application but still do the same task above:

```rust
fn system_b(windows: Query<Entity, With<&Window>>,
            window_ids: Local<Vec<Entity>>) {
    window_ids.clear();
    window_ids.extend(windows.iter());
}
```
You should also be wary of strings. Any call to `format!` allocates a string.
```rust
fn system_c(windows: Query<Entity, With<&Window>>) {
    for id in &windows {
        let s = format!("ID is {}", id);
    }
}
```
You can use `Local<String>` as I did before but for this string:

```rust
fn system_d(windows: Query<Entity, With<&Window>>,
            mut s: Local<String>) {
    for id in &windows {
        s.clear()
        let _ = write!(s, "ID is {}", id);
    }
}
```

But maybe you're not even in a system. Maybe you're in some other Rust code, then you can still minimize your allocations doing something like this:

```rust
fn deep_dark_code(...) {
    let mut s = String::new();
    for id in &windows {
        s.clear();
        let _ = write!(s, "ID is {}", id);
    }
}
```
6. Consider using an event if your enum has a no-operation variant.

We had a `RelayoutWindowPhase` which is used to move windows to screens and change their dimensions. It also had an `Idle` variant. `Idle` did nothing. In such cases it's probably the case that you want to fire an event to have it do something, and the idle case is the majority of the time where no event is fired.

7. Prefer field access to bare accessory methods.

It was considered good practice in OO to always shield access to fields via a method or property accessor.

```rust
struct A {
    a: usize,
    b: u32,
}

impl A {
    fn a(&self) -> usize {
        self.a
    }

    fn a_mut(&mut self) -> &mut usize {
        &mut self.a
    }
}
```
If you have bare accessors like the above, it is preferred to increase the visibility of your fields to `pub(crate)` or `pub` and manipulate the fields directly. It's clearer in the code what's happening. It's more performant. Many things that OO accessors aimed to guard against can't happen in Rust:

a. No one can stick `NULL` where some other value ought to be.
b. No one can write into your value unless they have a `&mut` or owned value.

One exception to this preference is when implementing traits, which cannot express field constraints.

8. Prefer .chain() to many .before() and .after() constraints in scheduling.

If you have systems that look like this:
```rust
app
    .add_systems(Update, a.before(b))
    .add_systems(Update, b.before(c))
    .add_systems(Update, c.before(d))
    .add_systems(Update, d.before(e));
```
Consider using a chain instead.
```rust
app
    .add_systems(Update, (a,
                          b,
                          c,
                          d).chain());
```

* * *

That's the end of my refactoring synopsis. There are things left to do. I have no doubt regressed some of the functionality, for which I am sorry, but I think the benefits of what I've done above will ease the struggle. 

* The plus button on the secondary window panels work, but panel plus button has ceased to work on the main window.
* `WindowRelayout::UpdateDescriptors` event needs to be fired by someone.
* The `WindowState::descriptor.path` needs to be set when its known for the primary window and should be used as its source of truth for the KDL.

### DIFF
--- a/libs/elodin-editor/src/ui/mod.rs
+++ b/libs/elodin-editor/src/ui/mod.rs
@@ -53,9 +53,7 @@ const DEFAULT_PRESENT_MODE: PresentMode = PresentMode::Fifo;
 
 #[cfg(target_os = "linux")]
 mod platform {
-    use std::time::Duration;
     pub const LINUX_MULTI_WINDOW: bool = true;
-    pub const SCREEN_RELAYOUT_TIMEOUT: Duration = Duration::from_millis(2000);
     pub const PRIMARY_ORDER_OFFSET: isize = 0;
 }
 


x46085:
Super fun PR to read, learned a bunch more about Bevy! Good to see the commitment to ideal architecture.

PR #334, opened 2025-12-03
feat(aleph) - add option for selecting an alternate kernel configuration

labrnth:
aleph (kernel) - add an option `aleph.kernel.source`, where the user can specify an option specifying to a branch of
aleph-orin-baseboard-kernel-5-10 corresponding to different kernel configurations. For now, two options are offererd:
- `default` : the existing config
- `no_otg` : a config where usbc-1 is set to a host port (not otg)

Note1: It looks like there are a lot of changes in  `aleph/modules/hardware.nix`, but they are mostly indentation changes. I wish the diff were more intelligent in conveying this.
Note2: if you enable `no_otg`, you will also need to disable the `usb-eth.nix` module, as the usb0 device will not be available under the `no_otg` configuration.

### DIFF
--- a/aleph/kernel/default.nix
+++ b/aleph/kernel/default.nix
@@ -8,149 +8,169 @@
   structuredExtraConfig ? {},
   argsOverride ? {},
   buildLinux,
+  kernelSource ? "default",
   ...
-} @ args:
-buildLinux (args
-  // {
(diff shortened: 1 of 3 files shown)


labrnth:
Added sample configuration entry in `template/flake.nix` and force-pushed.

x46085:
Need to run `alejandra`

labrnth:
Fixed linting, pushing ahead. Thank you!

PR #335, opened 2025-12-04
fix(aleph) - fix default elodin-db boot service

labrnth:
aleph(elodin-db) - Commit d3921b973b3faf65b3e01389426fd91e0383ca99 changed the elodin-db service to add a timestamp to the database name. This was useful in the when the default database was started on boot, but it had the unintended side effect of appending a time stamp to every custom invocation of the elodin-db template service.

This commit introduces three modifications:
1. Revert the change to the template service, so that the timestamp is not applied to custom invocations
2. Create a one-shot deafult service which invokes the template service at boot with a with a default name and a timestamp if desired
3. Add an option to choose whether or not this 'autostart' service is created in the first place.

### DIFF
--- a/aleph/modules/elodin-db.nix
+++ b/aleph/modules/elodin-db.nix
@@ -6,16 +6,6 @@
 }: let
   elodin-db = pkgs.elodin-db;
   cfg = config.services.elodin-db;
-  elodin-db-wrapper = pkgs.writeShellScriptBin "elodin-db-wrapper" ''
-    DB_NAME="$1"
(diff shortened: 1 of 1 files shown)


labrnth:
Tested all of the options this time :disappointed:

x46085 (approved):

labrnth:
Fixed formatting. Thanks!

labrnth:
Formatting fixed and rebased onto main. Thank you!

PR #339, opened 2025-12-05
fix-error-log: avoid double logging subscriber

zxq82lm:
## Purpose
- Fixing 'ERROR Could not set global logger and tracing subscriber as they are already set. Consider disabling LogPlugin.'


x46085 (approved):
beautiful, ty for fixing!

PR #349, opened 2025-12-09
feat: reset cameras, new emissivity kdl-field, new lock kdl-field (graph) - fix: prevent non-expected drag on click a panel

zxq82lm:
## Reset cameras
Command palette > Reset cameras > All viewports:
image
After the reset:
image

## emissivity=
```kdl
object_3d a.world_pos {
    sphere radius=0.2 emissivity=1.0 {
        color yellow
    }
}
object_3d b.world_pos {
    sphere radius=0.2 emissivity=1.0 {
        color pink
    }
}
object_3d c.world_pos {
    sphere radius=0.2 emissivity=1.0 {
        color cyan
    }
}
```

Screenshot 2025-12-10 at 08 34 16

## lock=  (graph)
```kdl
graph "drone.attitude_target[0], drone.attitude_target[1], drone.attitude_target[2], drone.attitude_target[3], drone.world_pos.q0, drone.world_pos.q1, drone.world_pos.q2, drone.world_pos.q3" name="World Pos" lock=#true
```

## fix: non-expected drag
- Now dragging is only triggered once the mouse cursor has moved more than 12 pixels.

### DIFF
--- a/libs/elodin-editor/src/ui/command_palette/palette_items.rs
+++ b/libs/elodin-editor/src/ui/command_palette/palette_items.rs
@@ -529,6 +529,65 @@ fn toggle_body_axes() -> PaletteItem {
     })
 }
 
+fn reset_cameras() -> PaletteItem {
+    PaletteItem::new(
+        "Reset Cameras",
+        VIEWPORT_LABEL,
+        |_: In<String>, window_states: Query<(&tiles::WindowState, &tiles::WindowId)>| {
+            let entries = gather_viewport_entries(&window_states);
+            if entries.is_empty() {
+                return PalettePage::new(vec![PaletteItem::new(
+                    "No viewports available",
+                    VIEWPORT_LABEL,
+                    |_: In<String>| PaletteEvent::Exit,
+                )])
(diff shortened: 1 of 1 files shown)


libs/elodin-editor/src/ui/command_palette/palette_items.rs
```diff
@@ -1513,6 +1583,64 @@ impl Default for PalettePage {
                 },
             ),
             toggle_body_axes(),
+            PaletteItem::new(
```
anderspitman: Can we move this into a function (like`save_schematic`)? Especially since it's pretty long.

shanecelis: I think Benoît asked me about lumping these together and I said, yes, so that's on me, but I agree in the future we ought to separate them.

zxq82lm: Thank you, I just added a reset_cameras function in command palette.

anderspitman (approved):

PR #352, opened 2025-12-10
Add support for configurable reference frames to Bevy.

shanecelis:
This PR is built on top of #328. 

# TODO

Here are the remaining TODOs:

- [x] Rebase onto main, which requires updating to Bevy 0.17.
- [x] Integrate it into Elodin position syncing.
  - [x] Handle camera
  - [x] Handle 3d object
  - [x] Handle vector arrow
  - [x] Handle line 
  - [x] Handle grid colors
- [x] Handle the axes widget depending on the chosen frame.
- [x] Update documentation

# API

The API at runtime is centered around these two components.

```rust
#[derive(Component)]
pub struct GeoPosition(pub GeoFrame, pub DVec3);
#[derive(Component)]
pub struct GeoRotation(pub GeoFrame, pub DQuat);
```

```rust
commands.spawn(GeoPosition(GeoFrame::NED, DVec3::new(0.0, 1.0, 0.0)));
commands.spawn(GeoRotation(GeoFrame::NED, DQuat::IDENTITY);
```

In addition there is a `GeoContext` that contains the latitude, longitude, and altitude, which are all by default zero, and we currently don't have a way to specify them at runtime, but they do work and can be exercised in the bevy_geo_frames example binary.

# How to exercise these features

## Change the coordinate frame

Add "coordinate frame=NED" to the top of a KDL, and it will be globally be in NED mode. However, most any KDL element that accepts a position can also specify a "frame=NED" attribute. So the KDL can mix and match coordinate systems.

The cube axes will show the correct XYZ axes and color for each mode. It uses the viewport's frame if specified. For example here is the ECEF cube widget:

Screenshot_2026-03-26_at_2 19 35_PM

Here is the NED cube widget:

Screenshot_2026-03-23_at_5 55 53_PM

## Ball example

The ball example has its regular ENU simulation in sim.py as before, but it also has a sim_ned.py that you can switch to by (un)commenting some code in ball/main.py. 

# Design considerations

If no frame is specified, ENU is used as the default, so GeoPosition and GeoRotation are always in the code path. This was done to avoid having two similar but different code paths: declared ENU using GeoPosition and undeclared using the old code. This ENU as default is controlled by the cargo feature "enu_default" in the bevy_geo_frames crate.

# Future Work

The next things to do in this vein would be the following:
- Add a means of specifying latitude, longitude, and altitude when specifying the coordinate frame in the KDL.
- Provide a means of switching from the "Plane" view versus the "Planetary" view similar to the demo.
- Consider adding a Sun-centric coordinate system like HCI.
- Maybe adds tooltips for coordinate systems like ECEF that don't have a nice cardinal directions.
- Currently GeoRotation consumes Elodin's Bevy quaternion, which is doing its own ENU conversion. This may not play nicely with NED geo rotations but needs investigation.


shanecelis:
# Current 

Here's a video showing a demo of the little crate I wrote that handles this coordinate frame stuff. This shows the user toggling between three coordinate frames ENU and NED and ECEF and two presentation modes: plane, which is our standard view; and sphere, which demonstrates where the coordinate system with respect to the Earth, or in this case 10 m radius sphere. The 1x2x3 cuboid is rotating with respect to an ENU coordinate system on its Up axis. 

https://github.com/user-attachments/assets/d1b037cb-0158-4d99-8fe0-49c1903b4a0c

# Design

The design for the user can be summed up like this:
```rust
#[derive(Component)]
pub struct GeoPosition(pub GeoFrame, pub DVec3);
```

You specify your coordinate system with your coordinate, and the systems will sync then when they are changed.
```rust
commands.spawn(GeoPosition(GeoFrame::ENU, DVec3::new(0.0, 1.0, 0.0)));
```

You can also specify rotations, velocities, and angular velocities with respect to a coordinate frame.

shanecelis:
Moved TODOs to top-level.

libs/bevy_geo_frames/src/geo.rs
```diff
@@ -0,0 +1,1055 @@
...
+    /// Convert vector to Bevy.
+    pub fn to_bevy(&self, context: &GeoContext) -> DVec3 {
+        GeoFrame::bevy_R_(&self.0, context) * self.1
+    }
+
+    /// Convert vector from Bevy.
+    pub fn from_bevy(frame: &GeoFrame, v_bevy: impl Into<DVec3>, context: &GeoContext) -> Self {
+        let v = v_bevy.into();
+        let w = GeoFrame::bevy_R_(frame, context).transpose() * v;
+        GeoVelocity(*frame, GeoFrame::bevy_R_(frame, context).transpose() * w)
```
zxq82lm: This two lines (376+377) should not be removed?
GeoVelocity::from_bevy applies the transpose twice, but w seems already right.

libs/bevy_geo_frames/src/geo.rs
```diff
@@ -0,0 +1,1055 @@
...
+    let dt = time.delta_secs_f64();
+    if dt == 0.0 {
+        return;
+    }
+
+    for (mut geo_rot, ang) in &mut q {
+        let rot0_R_ang0 = geo_rot.0._R_(&ang.0, &ctx);
+        if geo_rot.0 != ang.0 {
+            // We're punting.
+            continue;
```
zxq82lm: integrate_geo_orientation could silently ignore cross-frame rotations.
Convert is already done (convert matrix between 2 frames -> rot0_R_ang0), but the "continue;" seems to shortcut it.

libs/bevy_geo_frames/src/geo.rs
shanecelis: Good catch. Will fix.

libs/bevy_geo_frames/src/geo.rs
shanecelis: Good find. The `GeoVelocity` and `GeoAngularVelocity` are there in the crate but not used in Elodin itself. It's good to fix though.

zxq82lm (approved):
LGTM

PR #354, opened 2025-12-12
[TEST] refactor(ui): per-window sidebars — Hierarchy (left) and Inspector (right).

zxq82lm:
## Content
- Each window has a maskable Hierarchy sidebar on the right and a maskable Inspector sidebar on the right.
- Even when there is only a single pane, its parent tab remains visible.
- A single click on a relevant pane/tile title unmasks the Inspector in the current window (if it is masked) and shows that pane/tile’s content.
- Clicking a line in the Hierarchy continues to show the relevant Inspector content.
- All pane and tab titles are editable and persistable (double-click titles to edit).
- doc: New documentation Command Palette.
- Command Palette cleanup (removing Inspector & Hierarchy).
- Cleanup of unused or unimplemented modals.
- Auto-arrange windows per screen when no kdl-rect= is defined.

image


zxq82lm:
Include the approach for implementing the Hierarchy and Inspector sidebars in the codebase. The other aspects discussed here will be handled in separate PRs, and the sidebars themselves in a dedicated PR.

PR #360, opened 2025-12-17
Update bevy egui

anderspitman:
Update bevy_egui to 0.35.1

This is in preparation for updating to bevy 0.17. bevy_egui 0.35.0
introduced breaking changes where the egui Context is now attached
to cameras instead of windows, and only the first camera.
(see https://github.com/vladbat00/bevy_egui/releases/tag/v0.35.0)

The primary trickiness comes from supporting multiple windows. I
tried to use the recommended bevy_egui way, but it was not a good
fit because we dynamically spawn our windows.
(see https://github.com/vladbat00/bevy_egui/issues/448)

Finally decided to manage the egui contexts manually.

A few things to look out for:

* The egui inspector overlay is now in its own window
* UI is not manually ordered above everything else
* We are no longer relying on any of bevy_egui's automatic
  functionality
* Removed load_material_icons for the plot, because it was
  causing errors with the new egui stuff and as far as I can tell
  it's not being used there

---

> [!NOTE]
> Upgrade bevy_egui and migrate to camera-attached egui with manual per-window contexts, explicit UI camera ordering, and an inspector in its own window.
> 
> - **UI/egui integration**:
>   - Upgrade `bevy_egui` to `0.35.1` and `bevy-inspector-egui` to `0.32`.
>   - Switch to `EguiPlugin::default()` and disable auto primary context via `EguiGlobalSettings`.
>   - Migrate to camera-attached `EguiContext`: manually create/attach `EguiContext` and `Camera2d` to the primary and secondary window entities (`RenderTarget::Window(WindowRef::Entity(...))`).
>   - Introduce explicit UI ordering with `UI_ORDER_BASE` and set camera `order` for primary/secondary windows; adjust systems to run `sync_windows` in `PreUpdate` before `EguiPreUpdateSet::BeginPass`.
> - **Windows/Startup**:
>   - Secondary window spawn now includes `EguiContext` + `Camera2d`; primary startup window similarly updated.
>   - Move various window/camera viewport/order calculations to reflect new egui setup.
> - **Inspector** (feature `inspector`):
>   - Replace quick plugin with a dedicated “World Inspector” window using its own `EguiContext` + `Camera2d`; add `setup_egui_inspector` and `run_egui_inspector` systems.
> - **Plot/UI cleanup**:
>   - Remove `load_material_icons` from plots (no longer needed/compatible).
> - **Dependencies**:
>   - Update `Cargo.toml` and lockfile to new egui-related versions; minor additions like `EguiPreUpdateSet` usage and new Bevy window/camera refs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1461f0cb32f09935322f194c17276350d52e4ae2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.buildkite/pipeline.py
+++ b/.buildkite/pipeline.py
@@ -116,23 +116,22 @@ test_steps = [
     group(
         name=":nix: aleph-os",
         steps=[
-            step(
-                label=":nix: toplevel",
-                key="toplevel",
-                command=["cd aleph", "nix build --accept-flake-config .#toplevel"],
-                agents={"queue": "nixos-arm-aws"},
-                if_="false", # disable temporarily
-            ),
-            step(
-                label=":nix: sdimage",
-                key="sdimage",
-                command=[
-                    "cd aleph",
-                    "nix build --accept-flake-config .#sdimage",
-                ],
-                agents={"queue": "nixos-arm-aws"},
-                if_="false", # disable temporarily
-            ),
+            # # Disable temporarily
+            # step(
+            #     label=":nix: toplevel",
+            #     key="toplevel",
+            #     command=["cd aleph", "nix build --accept-flake-config .#toplevel"],
+            #     agents={"queue": "nixos-arm-aws"},
+            # ),
+            # step(
+            #     label=":nix: sdimage",
+            #     key="sdimage",
+            #     command=[
+            #         "cd aleph",
+            #         "nix build --accept-flake-config .#sdimage",
+            #     ],
+            #     agents={"queue": "nixos-arm-aws"},
+            # ),
             step(
                 label=":nix: flash-uefi",
                 key="flash-uefi",


anderspitman:
Another thing to note is that I think our current camera ordering is a bit brittle. I keep running into conflicts. For example, when I merged main into my branch, the arrow labels started rendering on top of the UI, because a new order had been added for the arrows since my last update. Basically we're going to keep track of every order in our app to avoid conflicts.

Also, having hard coded ranges puts a discrete limit on the number of each type we can have. Maybe this doesn't matter for practical purposes if the ranges are big enough, but it maybe it will in some circumstances.

It would be nice to figure out another way to solve this, but I don't have a solution off the top of my head.

shanecelis (approved):
Looks good and works well.

PR #362, opened 2025-12-18
feat: Allow absolute positioning of arrows.

shanecelis:
# Problem

Dynamic vector labels can move wildly on the screen, e.g., the velocity arrow in the ball demo. It can be hard to read the names of a vector arrow label when it's attached to the tip of one of these wild dynamic arrows.

# Solution

Allow the user to specify an absolute position for the vector_arrow in meters. This stabilizes the labels so that they do not move so wildly.

The users specify a absolute position to, say, 0.1 meters by putting this in the KDL:

```kdl
vector_arrow ... label_position="0.1m" {
```
All the prior proportionate specifying of arrows is retained.

https://github.com/user-attachments/assets/00a10a7c-4efc-43f2-a045-8e9f41cf9f40

### DIFF
--- a/libs/impeller2/wkt/src/gui.rs
+++ b/libs/impeller2/wkt/src/gui.rs
@@ -345,7 +345,6 @@ pub enum LabelPosition {
     Absolute(f32),
 }
 
-
 #[derive(Clone, Copy, Debug, PartialEq)]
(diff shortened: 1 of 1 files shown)


x46085:
I would expect to see a docs update here for this:
https://docs.elodin.systems/reference/schematic/#vector-arrow

shanecelis:
Right you are. Added to docs.

PR #368, opened 2025-12-18
Use same theme for all windows

### LINKED ISSUE #358: Primary and secondary windows are using different fonts
Here's an example from the plots

Image

anderspitman:
* Set themes on secondary windows
* No longer setting theme on every frame. Now just at egui context creation.
* Centralized creation of egui contexts behind a function, which lets us do setup like theme setting

fixes #358


anderspitman:
Weird I wonder how that slipped through

anderspitman:
Ah because it's conditional on the inspector feature

PR #372, opened 2025-12-19
UI Window Refactor + Theme/Scheme Management

zxq82lm:
## Summary
- Extracted window management into a dedicated `ui/window` module and trimmed `ui/mod.rs`.
- Added theme/scheme presets with KDL support, persistence, and runtime switching.
- Updated palette + timeline UI to reflect active scheme and added a Dark/Light toggle.
- Fixed Create Window empty layout sizing in secondary windows.
- Documented scheme presets and updated schematic reference + example.

## Refactor: ui/window module
- New `libs/elodin-editor/src/ui/window/{default,placement,spawn}.rs` and re-exports in `libs/elodin-editor/src/ui/window/mod.rs`.
- Window defaults centralized via `base_window`, `default_*`, and `window_theme_for_mode`.
- Window placement/relayout logic moved out of `ui/mod.rs`, including macOS physical rect placement and screen detection.
- Startup/editor windows now use shared defaults; explicit `WindowTheme` in `libs/elodin-editor/src/lib.rs` is removed.

## Feature: theme/scheme presets + registry
- `ThemeConfig` added to schematics; KDL `theme` supports `mode` and `scheme`.
- Schematic load applies theme to the session; save writes it back to KDL.
- `WindowDescriptor` now stores `mode` so secondary windows inherit the session mode (with per-secondary overrides).
- Preset registry merges built-ins with user JSON presets from `color_schemes/` (assets and data dir); user presets override built-ins on name collisions.
- Selection persistence stores `{scheme, mode, colors}` in `color_scheme.json`, with fallback for legacy files.

### KDL example
```kdl
theme mode="light" scheme="eggplant"
```

### Built-in presets
- `default` (dark + light)
- `eggplant` (dark + light)
- `catppuccini-macchiato` (dark + light)
- `catppuccini-mocha` (dark only)
- `catppuccini-latte` (light-preferred; dark uses mocha)
- `matrix` (dark + light)

## UI updates
- Command Palette: new items `Set Color Scheme` and `Set Color Scheme Mode` (unavailable variants are flagged).
- Command Palette search input now uses scheme colors (fixes dark-on-dark text).
- Timeline controls add a Dark/Light toggle to the right of the transport buttons.
- Arrow label text color always uses `text_primary` from the active scheme (ignores arrow-defined label color).

## Docs and examples
- New reference: `docs/public/content/reference/color-schemes.md`.
- Updated schematic reference for `theme` and preset registry behavior: `docs/public/content/reference/schematic.md`.
- Drone example sets an explicit theme: `examples/drone/sim.py`.

---

https://github.com/user-attachments/assets/8348d814-6672-4669-9c9e-1e5755ee9ced

### DIFF
--- a/docs/public/content/reference/color-schemes.md
+++ b/docs/public/content/reference/color-schemes.md
@@ -12,16 +12,38 @@ icon = ""
 order = 7
 +++
 
-## Presets
+## Overview
 
-- Built-ins: `default`, `eggplant`, `catppuccini-macchiato`, `catppuccini-mocha`, `catppuccini-latte`, and `matrix` ship with the app; most have dark and light variants, while mocha is dark-only and latte is light-preferred.
-- Registry merge: user presets are loaded from disk and merged with the built-ins; if a user preset reuses a built-in name, the user version replaces it.
-- Modes: dark is required; light is optional. When a light file is missing, the UI stays in dark mode and the palette entry shows the mode as unavailable.
+Color schemes define the UI palette. Each preset has a required dark variant and an optional light
+variant. Built-ins ship with the app and do not depend on your asset directory.
+
+## Built-in presets
+
+- `default` (dark + light)
+- `eggplant` (dark + light)
+- `catppuccini-macchiato` (dark + light)
+- `catppuccini-mocha` (dark only)
+- `catppuccini-latte` (light-preferred; dark uses mocha)
+- `matrix` (dark + light)
+
+## Registry and precedence
+
+- Built-ins are always available.
+- User presets are loaded from disk and merged with the built-ins.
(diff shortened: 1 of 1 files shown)


x46085:
Amazing! A dream realized. 

A couple tweaks I’d like to see:
- for default schemes, I wouldn’t want them to use the asset folder, or at least avoid the problem I have where I open Elodin all the time in random places and lose the assets. It would be ideal to have a code-based set of assets that are always built into Elodin and available regardless of asset directory settings. The default scheme and other @Teegurks curated themes would be there always.
- I really want a dark/light mode toggle button. The command palette is good and should keep it, but it’s not nearly as discoverable as this feature should be. Maybe to the right of the pause / play time controls?

PR #373, opened 2025-12-20
Default Data Overview Panel

x46085:
image

This branch introduces a Data Overview panel for the Elodin Editor that serves as the default view when connecting to a database without a predefined schematic. The panel displays all database components as rows with miniature sparkline graphs showing data trends over time, helping users quickly identify gaps or anomalies in their telemetry data. To handle large datasets without crashing, the system uses a two-phase query approach that first retrieves row counts, then applies server-side SQL downsampling when data volume exceeds a threshold. The panel supports interactive zoom and pan controls, and periodically refreshes to show live data during running simulations.

A significant fix addresses custom timestamp source fields in Elodin-DB. Users can define custom clock fields as timestamp sources for data playback, but these raw clock values were polluting timeline calculations, causing the editor to display incorrect 30-day ranges when actual data spanned only minutes. The solution was implemented at the database level where VTable analysis now identifies timestamp source fields and marks them in component metadata. The database automatically excludes these components from time range calculations, making the behavior transparent to all clients.

The branch also fixes UI consistency issues with the egui_tiles library. The primary window was missing tab bars that appeared correctly in secondary windows, traced to how egui_tiles only renders tab bars for nested Tabs containers. The tile insertion logic now ensures panes are placed within properly nested container structures, and simplification options prevent pruning of single-child tabs containers.

This is the full testing flow:

https://github.com/user-attachments/assets/f2f5a4ed-a601-4d7f-8026-cd78b496399d

---

> [!NOTE]
> Adds a Data Overview panel (default view) with sparkline timelines and two-phase SQL querying, and updates DB to mark/exclude timestamp-source fields from time range calculations.
> 
> - **Editor (UI/UX)**:
>   - **New `Data Overview` panel**: Timeline of all components with per-field sparklines, zoom/pan, periodic refresh; two-phase SQL (range/count → downsampled data). Integrated as default view when no schematic, command palette action, and tile type (`Pane::DataOverview`).
>   - **Query orchestration**: `ComponentTimeRanges` cache and `trigger_time_range_queries` system; clears cache on new connections.
>   - **Tiles/Layouts**: Ensure nested Tabs so tab bars render; robust `insert_tile` handling and prevent pruning single-child tabs.
> - **Database**:
>   - **Timestamp-source detection**: Analyze `VTable` ops to find timestamp source byte ranges; mark overlapping fields as `_is_timestamp_source` in `ComponentMetadata` and persist.
>   - **Time range correctness**: Exclude timestamp-source components from DB-wide time range calculations when opening DB and when inserting components.
> - **Shared/Meta**:
>   - `ComponentMetadata` gains `is_timestamp_source`/`set_timestamp_source` helpers.
>   - Editor deps: add `convert_case` for table-name derivation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1a982d2c432f5b910dba1a593a667a358714d2f8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/libs/elodin-editor/src/ui/data_overview/mod.rs
+++ b/libs/elodin-editor/src/ui/data_overview/mod.rs
@@ -430,7 +430,8 @@ impl WidgetSystem for DataOverviewWidget<'_, '_> {
             .unwrap_or_else(|| (params.selected_range.0.start, params.selected_range.0.end));
 
         // Add a small margin (5%) to the data range for better visualization
-        let base_span = (data_max.0 - data_min.0).max(1_000_000) as f64; // minimum 1 second
+        let span_us = data_max.0.saturating_sub(data_min.0);
+        let base_span = span_us.max(1_000_000) as f64; // minimum 1 second
         let margin = (base_span * 0.05) as i64;
         let display_data_min = Timestamp(data_min.0.saturating_sub(margin));
 
@@ -441,8 +442,8 @@ impl WidgetSystem for DataOverviewWidget<'_, '_> {
         let max_pan = (base_span - zoomed_span).max(0.0) as i64;
         pane.pan_offset_us = pane.pan_offset_us.clamp(0, max_pan);
 
-        let display_start = Timestamp(display_data_min.0 + pane.pan_offset_us);
-        let display_end = Timestamp(display_start.0 + zoomed_span as i64);
+        let display_start = Timestamp(display_data_min.0.saturating_add(pane.pan_offset_us));
+        let display_end = Timestamp(display_start.0.saturating_add(zoomed_span as i64));
 
         let timeline_width = (available_rect.width() - LABEL_WIDTH).max(100.0);
         let time_span = (display_end.0 - display_start.0).max(1) as f64;


libs/elodin-editor/src/ui/data_overview/mod.rs
```diff
@@ -0,0 +1,1114 @@
...
+                        None
+                    } else {
+                        Some(vals.value(i) as f64)
+                    }
+                })
+                .collect()
+        })
+    };
+
+    values.map(|v| vec![v])
```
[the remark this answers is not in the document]
x46085: Fixed

x46085:
@zxq82lm when you get the light / dark PR landed, I could use your help updating this branch with your changes and then testing it all is working well. Your PR may include some fixes I also made, so having you take this would be helpful.

zxq82lm:
> @zxq82lm when you get the light / dark PR landed, I could use your help updating this branch with your changes and then testing it all is working well. Your PR may include some fixes I also made, so having you take this would be helpful.
---
@x46085
Yep, no problem. We can agree that I’ll take care of rebasing this branch on main?

libs/elodin-editor/src/ui/data_overview/mod.rs
```diff
@@ -0,0 +1,1122 @@
...
+
+        // Apply zoom factor (1.0 = fit all data, >1.0 = zoomed in)
+        let zoomed_span = base_span / pane.zoom_factor.max(0.01) as f64;
+
+        // Apply pan offset (clamped to prevent scrolling beyond data)
+        let max_pan = (base_span - zoomed_span).max(0.0) as i64;
+        pane.pan_offset_us = pane.pan_offset_us.clamp(0, max_pan);
+
+        let display_start = Timestamp(display_data_min.0 + pane.pan_offset_us);
+        let display_end = Timestamp(display_start.0 + zoomed_span as i64);
```
zxq82lm: See related explanation above (Files changed) or below (Conversation):

```
let display_start = Timestamp(display_data_min.0.saturating_add(pane.pan_offset_us));
let display_end = Timestamp(display_start.0.saturating_add(zoomed_span as i64));
```

libs/elodin-editor/src/ui/data_overview/mod.rs
```diff
@@ -0,0 +1,1122 @@
...
+        }
+
+        // Use cached data range, falling back to selected_range if no data yet
+        let (data_min, data_max) = pane
+            .cached_data_range
+            .or(data_time_range)
+            .unwrap_or_else(|| (params.selected_range.0.start, params.selected_range.0.end));
+
+        // Add a small margin (5%) to the data range for better visualization
+        let base_span = (data_max.0 - data_min.0).max(1_000_000) as f64; // minimum 1 second
```
zxq82lm: When loading an example (drone or otherwise), there is a schematic loaded from a path. The UI is then built using the panels defined by the schematic.
- The Data Overview panel computes a "time span" in order to render the timeline.
- At the very beginning, there is still no data in the database, so there is no real interval (no meaningful min/max).
- It then falls back to the default time selection values: Timestamp(i64::MIN) → Timestamp(i64::MAX).
- When it tries to compute max - min, the subtraction overflows (i64::MAX - i64::MIN) and panics.

I tested locally this patch successfully (not sure it is optimal):
```
let span_us = data_max.0.saturating_sub(data_min.0);
let base_span = span_us.max(1_000_000) as f64; // minimum 1 second 
```
And this two lines below (Files changed) or above (Conversation):
```
let display_start = Timestamp(display_data_min.0.saturating_add(pane.pan_offset_us));
let display_end = Timestamp(display_start.0.saturating_add(zoomed_span as i64));
```
---
The idea of the patch:
- saturating_sub performs a "safe" subtraction: if the operation exceeds the numeric bounds, it saturates instead of panicking.
- Then max(1_000_000) enforces a minimum of 1 second, even if the computed span is 0.

PR #382, opened 2025-12-28
Always-on hierarchy/inspector sidebars per window.

zxq82lm:
> Merged through https://github.com/elodin-sys/elodin/pull/385

## Content
- Always-on hierarchy/inspector sidebars per window:
   - Sidebars can be masked/unmasked by clicking their gutters.
   - When unmasked, sidebars are resizable via the gutter slider.
   - Clicking a pane (e.g. a Graph) unmasks the Inspector and shows the matching content.
   - Clicking a Hierarchy row, updates the Inspector to the corresponding entity.

## Following PRs:
- https://github.com/elodin-sys/elodin/pull/384 (based on current 382)
   - https://github.com/elodin-sys/elodin/pull/385 (based on 384)
   
---
- Video (on a single window):

https://github.com/user-attachments/assets/04c32722-90b0-46d0-aa31-444e70104ee2

### DIFF
--- a/docs/public/content/reference/command-palette.md
+++ b/docs/public/content/reference/command-palette.md
@@ -35,6 +35,8 @@ order = 8
 - `Create Data Overview`: add a data overview pane.
 - `Create 3D Object`: prompt for an EQL expression, pick a mesh type (GLTF or primitives), then
   optionally enter dimensions and color.
+- `Hierarchy` and `Inspector` are built-in sidebars; they are always present and do not appear in
+  the command palette.
 
 ## Viewport
 
--- a/docs/public/content/reference/command-palette.md
+++ b/docs/public/content/reference/command-palette.md
@@ -30,8 +30,6 @@ order = 8
 - `Create Action`: prompt for a label, then choose a `send_msg` preset or enter a custom Lua
   command to create an action pane.
 - `Create Video Stream`: prompt for the message name, then create a video stream pane.
-- `Create Hierarchy`: add a hierarchy pane.
-- `Create Inspector`: add an inspector pane.
 - `Create Schematic Tree`: add a schematic tree pane.
 - `Create Dashboard`: add a dashboard pane.
 - `Create Data Overview`: add a data overview pane.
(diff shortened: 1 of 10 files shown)


x46085:
When making visual / UI / UX changes, I find it very useful to drop a little video, like 10-20 seconds, into the description to show the updated behavior.

x46085 (approved):
Functionally this looks great.

I'd like to grab some eyeballs from @Teegurks on this soon to think about the styling of these bars. They are a bit thick and the color is a bit disjointed from the rest of the app's style. I'm sure Tom will have some tweaks that could be applied to make the bars more evocative of "click me" or "I'm hiding something", be thinner and match the app's style and feel. I'll ask him when we chat this week.

zxq82lm:
> Functionally this looks great.
> 
> I'd like to grab some eyeballs from @Teegurks on this soon to think about the styling of these bars. They are a bit thick and the color is a bit disjointed from the rest of the app's style. I'm sure Tom will have some tweaks that could be applied to make the bars more evocative of "click me" or "I'm hiding something", be thinner and match the app's style and feel. I'll ask him when we chat this week.
---
Thanks a lot @x46085 & @Teegurks.
- For now, the gutters’ gray is arbitrary and not wired into the theme system (scheme/mode), so it stays the same across all schemes in both dark and light mode.
- This also gives me time to exercise the sidebar behavior across all examples, including the single-window Data Overview case (so far I’ve only tested drone and ball).

zxq82lm:
@anderspitman @shanecelis @x46085:
Assume two windows, V1 and V2:
- You click on Graph G1 in V1, so V1’s Inspector becomes unmasked and shows Graph G1.
- You click on Graph G2 in V2, so V2’s Inspector becomes unmasked and shows Graph G2. In the current version, V1’s Inspector (still unmasked) is also showing Graph G2.

Should we change this behavior? In this example, should V1’s Inspector continue to display the data for Graph G1?
In other words: an unmasked Inspector should remain scoped to its own window and keep showing the last selection made in that window; it should not have its displayed content updated by Inspector activity in other windows?

x46085:
> Should we change this behavior? In this example, should V1’s Inspector continue to display the data for Graph G1? In other words: an unmasked Inspector should remain scoped to its own window and keep showing the last selection made in that window; it should not have its displayed content updated by Inspector activity in other windows?

Yes, since the sidebars exist in every window, their behavior should be related only to the window they are in.

PR #395, opened 2026-01-07
fix: data overview default display when no schematic.

zxq82lm:
## Issue
- The default display of Data Overview when reading an Elodin DB without a Schematic is not correct (identified by 
@x46085, thanks).

image

---

## Testing
- Execute editor rc-jet example + Command Palette > Save DB (`rc-jet.db`).
- `cargo run --bin elodin-db --manifest-path /elodin/libs/db/Cargo.toml -- run 127.0.0.1:2240 /rc-jet.db` (terminal 1)
- `cargo run --bin elodin --manifest-path /elodin/apps/elodin/Cargo.toml -- editor 127.0.0.1:2240` (terminal 2)

image

### DIFF
--- a/libs/elodin-editor/src/ui/schematic/load.rs
+++ b/libs/elodin-editor/src/ui/schematic/load.rs
@@ -812,13 +812,14 @@ impl LoadSchematicParams<'_, '_> {
 
         let target_id = {
             let tree = &window_state.tile_state.tree;
-            tree.root().and_then(|root_id| match tree.tiles.get(root_id) {
-                Some(Tile::Container(Container::Linear(linear))) => {
-                    let center_idx = linear.children.len() / 2;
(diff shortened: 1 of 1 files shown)


x46085:
Thanks for jumping on this!

Is it possible to restore the tab on top as well? You can see it shows as cut off at the top

Previously it would load and automatically create itself into a new panel with a tab so that the plus icon UI was present and you could add new panels as expected.

zxq82lm:
> Thanks for jumping on this!
> 
> Is it possible to restore the tab on top as well? You can see it shows as cut off at the top
> 
> Previously it would load and automatically create itself into a new panel with a tab so that the plus icon UI was present and you could add new panels as expected.

Sure yes. I’ll handle it.

PR #398, opened 2026-01-07
Chore/bevy egui 0.36.0

anderspitman:
Mostly needed to figure out the new Popup API. We're actually only half using it here. I spent some time seeing about completely modernizing it, but ended up deciding to keep it as close as possible to the old code while removing deprecated code. We might consider revisiting and cleaning up a bit in the future. We might be able to remove some of the manual interaction checks.

Main concern with merging this is that it touched some of the dashboard code, which we're not currently able to test.


anderspitman:
Ok in the process of fixing clippy errors I actually cleaned up the popup usage quite a bit. I feel pretty good about it now.

zxq82lm (approved):
Everything seems to be working perfectly: great job!
I successfully tested the following UI components: 
- Inspector Graph
- Inspector widgets (including the color picker)
- Plot widget
- Schematic Tree
- Tiles layout
- Tiles sidebar
- Timeline controls
- QueryTable

x46085 (approved):

anderspitman:
Thanks @zxq82lm!

PR #399, opened 2026-01-09
Chore/bevy 0.17

anderspitman:


libs/elodin-editor/src/ui/command_palette/palette_items.rs
```diff
@@ -133,6 +133,10 @@ impl PalettePage {
         self.prompt = Some(prompt.to_string());
         self
     }
+
+    pub fn into_event(self) -> PaletteEvent {
```
anderspitman: All of the `into_event` calls are using this.

libs/elodin-editor/src/ui/command_palette/palette_items.rs
```diff
@@ -173,7 +177,7 @@ impl LabelSource {
     pub fn get(&mut self, world: &mut World, filter: &str) -> String {
         match self {
             LabelSource::Label(l) => l.clone(),
-            LabelSource::System(system) => system.run(filter, world),
+            LabelSource::System(system) => system.run(filter, world).expect("Missing label"),
```
anderspitman, line 180: This matches the old behavior, which would panic. We may want to handle explicitly.

See https://bevy.org/learn/migration-guides/0-16-to-0-17/#system-run-returns-result

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -798,8 +802,17 @@ fn sync_camera_grid_cell(
         }
     }
 }
-fn sync_hdr(hdr_enabled: Res<HdrEnabled>, mut query: Query<&mut Camera>) {
-    for mut cam in query.iter_mut() {
-        cam.hdr = hdr_enabled.0;
-    }
-}
+
```
anderspitman, line 805: HDR is now a component instead of Camera attribute.

See https://bevy.org/learn/migration-guides/0-16-to-0-17/#split-hdr-from-camera

libs/elodin-editor/src/ui/command_palette/palette_items.rs
```diff
@@ -133,6 +133,10 @@ impl PalettePage {
         self.prompt = Some(prompt.to_string());
         self
     }
+
```
anderspitman, line 136: All of the `into_event` calls are using the method below.

libs/elodin-editor/src/lib.rs
```diff
@@ -422,13 +429,14 @@ struct SetupTitlebar;
...
     windows: Query<Entity, Without<SetupTitlebar>>,
-    winit_windows: NonSend<bevy::winit::WinitWindows>,
     mut commands: Commands,
+    _non_send_marker: NonSendMarker,
 ) {
     use objc2::rc::Retained;
     use objc2::{ClassType, msg_send, msg_send_id};
     use objc2_app_kit::{NSColor, NSToolbar, NSWindow, NSWindowStyleMask, NSWindowToolbarStyle};
 
+    WINIT_WINDOWS.with_borrow(|winit_windows| {
```
anderspitman, line 439: `WINIT_WINDOWS` replaces the old `WinitWindows` resource, using a scoped function now. This indents a lot of code that otherwise hasn't changed. What I'm doing for now is not indenting properly. clippy will complain, but it makes a lot less code to review. Once review is done I'll push the indentation changes.

libs/elodin-editor/src/ui/plot/gpu.rs
```diff
@@ -316,6 +322,19 @@ impl FromWorld for LinePipeline {
     }
 }
 
+fn init_line_pipeline(
```
anderspitman, line 325: See https://bevy.org/learn/migration-guides/0-16-to-0-17/#many-render-resources-now-initialized-in-renderstartup

libs/elodin-editor/src/ui/startup_window.rs
```diff
@@ -261,17 +261,17 @@ impl RootWidgetSystem for StartupLayout<'_, '_> {
         let mut state = state.get_mut(world);
         let logo_full = state
             .contexts
-            .add_image(state.images.logo_full.clone_weak());
+            .add_image(EguiTextureHandle::Weak(state.images.logo_full.id()));
```
anderspitman, line 264: See https://bevy.org/learn/migration-guides/0-16-to-0-17/#handle-weak-has-been-replaced-by-handle-uuid

anderspitman (commented):
@shanecelis @zxq82lm I did a bit of a pre-review leaving some comments to explain a few things. Here's the bevy migration guide for reference:

https://bevy.org/learn/migration-guides/0-16-to-0-17/

Note that this is ready for review even though clippy is broken. See comment about `WINIT_WINDOWS` below.

libs/elodin-editor/src/ui/mod.rs
shanecelis: This is fine. It'll work but I feel like we could add some qualifiers like `Added<Camera>`, and checking whether `HdrEnabled.is_changed()` could be used to opt out early. But our camera number is a handful and it doesn't do anything in the loop that isn't necessary, so again it's fine.

libs/elodin-editor/src/lib.rs
```diff
@@ -526,18 +536,16 @@ fn handle_drag_resize(
             window.set_cursor(winit::window::CursorIcon::Default);
         }
     }
+    });
 }
 
 fn setup_window_icon(
-    _windows: Query<(Entity, &bevy::window::PrimaryWindow)>,
-    // this is load bearing, because it ensures that there is at
```
shanecelis, line 533: This is a comment I'm sad to see go just because it looks like it's trying to prevent an accidental removal. Is it still true? It looks like it's not true any longer. Is that correct?

shanecelis:
This looks pretty great, Anders! I did find a small issue on macOS. I fixed it and pushed a commit directly in this branch. I exercised the ball, drone, cube_sat, and three_body, and rocket examples with no problems. 

EDIT: Rocket works too.

libs/elodin-editor/src/lib.rs
anderspitman: Yeah I was torn on this too. Originally I had actually added an additional comment. When I looked closer I assumed that `!_windows.is_empty()` would be enough to ensure a window, but maybe not. It's also unclear why both parameters start with an underscore. We could defensively trust in Chesterton's Fence and keep using `WINIT_WINDOWS` here, or we could yolo and see if we ever have issues with random crashes on startup. I don't think that would be particularly hard to debug if it happened. I'm good either way, so your call.

anderspitman:
@shanecelis only had one followup question. Once that resolved I'll push a new commit that does nothing other than clippy/fmt.

libs/elodin-editor/src/lib.rs
anderspitman: I just realized my thinking `_windows` might be sufficient makes no sense, because it's not even used for macOS. I vote we defensively use WINIT_WINDOWS with an even more verbose comment.

anderspitman: @shanecelis I just pushed the closest port I can come up with for the old behavior, along with a comment. If it looks good to you and you wouldn't mind verifying that it doesn't break macOS, I think we can resolve this.

shanecelis:
It's working on macOS, but I am seeing this strange thing in the rocket example:

Screenshot 2026-01-15 at 10 01 15 AM

shanecelis:
Uh oh.

https://github.com/user-attachments/assets/d4cb1148-45ed-4ee4-8b60-4bfa8ab0bd1f

anderspitman:
Well that's weird. I'm seeing the panes not close properly on main as well, but the shared ID thing is unique to my branch.

How did you get that nice warning widget about the IDs?

shanecelis:
I didn't do anything special to get the warning. I'm just lucky I guess. :)

PR #410, opened 2026-01-13
CLI: 'elodin-db info' prints details (recording, time step, metadata, schematic) without requiring a running db.

zxq82lm:
## Content
Add `elodin-db info` command to inspect database configuration offline. Also tracks which elodin-db version created/opened each database.

## Usage
```
elodin-db info [PATH]
```

## Example
```
db_state: /tmp/version-test/db_state
version_created: 0.16.0-alpha.5
version_last_opened: 0.16.0-alpha.5
recording: true
default_stream_time_step: 8 ms
metadata:
  time.start_timestamp: 1769548515408923
```

### DIFF
--- a/libs/db/src/main.rs
+++ b/libs/db/src/main.rs
@@ -418,6 +418,15 @@ fn run_info(args: InfoArgs) -> miette::Result<()> {
         postcard::from_bytes(&bytes).map_err(|e| miette::miette!("decode error: {e}"))?;
 
     println!("db_state: {}", db_state_path.display());
+
+    // Display version information prominently
+    if let Some(version) = config.version_created() {
+        println!("version_created: {}", version);
+    }
+    if let Some(version) = config.version_last_opened() {
+        println!("version_last_opened: {}", version);
+    }
+
     println!("recording: {}", config.recording);
     println!(
         "default_stream_time_step: {}",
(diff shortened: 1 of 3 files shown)


x46085:
I would prefer anything like this to be packaged into the `elodin-db` binary, rather than other tools. Could make this one `elodin-db --info {PATH}`

zxq82lm:
> I would prefer anything like this to be packaged into the `elodin-db` binary, rather than other tools. Could make this one `elodin-db --info {PATH}`

@x46085 Good idea. I switch the PR in draft, and I do that.

shanecelis:
I think elodin-db has a nice subcommand convention, so maybe it ought to be `elodin-db info <PATH>`.

shanecelis:
One thing that would be really useful here is knowing what version of the database wrote the file. I'm assuming that information isn't actually stored anywhere yet.

libs/db/src/main.rs
```diff
@@ -221,5 +229,77 @@ async fn main() -> miette::Result<()> {
...
+fn run_info(args: InfoArgs) -> miette::Result<()> {
+    use impeller2_wkt::DbConfig;
+
+    let db_state_path = match args.path {
+        Some(path) => {
+            let state_path = path.join("db_state");
+            if state_path.exists() {
+                state_path
+            } else {
+                path
```
anderspitman, line 399: When provided with an invalid db path (`/tmp/.tmpSmpnoT` instead of `/tmp/.tmpSmpnoT/db`), I'm getting this error: `Error:   × failed to read /tmp/.tmpSmpnoT: Is a directory (os error 21)`

I think it's because of this line.

zxq82lm: Thanks! 
Done (commit: d17b4aa66689668938e7369762d29d23c2c8c135)

PR #412, opened 2026-01-14
time.start_timestamps management.

zxq82lm:
## Content
- Add `--start-timestamp` <micros> to elodin-db run and persist it in DbConfig.metadata.
- On DB open, read `time.start_timestamp` from metadata and keep it in memory as the DB time anchor.
- Change auto‑timestamp fallback to `start_timestamp + (now - db_start)` when a packet omits timestamps.
- Ensure `start_timestamp=0` yields monotonic‑from‑zero; default `start_timestamp=Timestamp::now()` preserves wall‑clock behavior.
- Surface `time.start_timestamp` in fix-timestamps.
- Add `--log-level` flag.

### DIFF
--- a/libs/db/src/lib.rs
+++ b/libs/db/src/lib.rs
@@ -285,8 +285,8 @@ pub struct DB {
     pub default_stream_time_step: AtomicU64,
     pub last_updated: AtomicCell<Timestamp>,
     pub earliest_timestamp: AtomicCell<Timestamp>,
-    // Wall-clock timestamp at the moment the auto timestamp baseline was set.
-    auto_timestamp_base_wall_clock: AtomicCell<Timestamp>,
+    // Wall-clock timestamp at the moment the DB start anchor was set.
+    db_start_wall_clock: AtomicCell<Timestamp>,
 }
 
 #[derive(Default)]
@@ -333,7 +333,7 @@ impl DB {
             default_stream_time_step,
             last_updated: AtomicCell::new(Timestamp(i64::MIN)),
             earliest_timestamp: AtomicCell::new(now),
-            auto_timestamp_base_wall_clock: AtomicCell::new(now),
+            db_start_wall_clock: AtomicCell::new(now),
         };
         db.save_db_state()?;
         Ok(db)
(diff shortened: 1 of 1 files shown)


x46085:
what is "auto_timestamp_base" ? it doesn't seem to be used for anything

also these changes seem to largely ignore the already exiting "db.set_earliest_timestamp(start_timestamp);", rather than incorporating them.

There are quite a few more lines changes here than needed, I was expecting to see:
1. use the existing "earliest_timestamp" field, rather than creating new "auto start" fields
2. use the logic of "earliest_timestamp + (now() - db_start)" for the place where component timestamps are injected

Should reduce the amount of code considerably. Feel free to clarify why more complexity is needed.

zxq82lm:
> what is "auto_timestamp_base" ? it doesn't seem to be used for anything
> 
> also these changes seem to largely ignore the already exiting "db.set_earliest_timestamp(start_timestamp);", rather than incorporating them.
> 
> There are quite a few more lines changes here than needed, I was expecting to see:
> 
> 1. use the existing "earliest_timestamp" field, rather than creating new "auto start" fields
> 2. use the logic of "earliest_timestamp + (now() - db_start)" for the place where component timestamps are injected
> 
> Should reduce the amount of code considerably. Feel free to clarify why more complexity is needed.

I did theses changes:
- `auto_timestamp_start` is removed; auto timestamps now use `earliest_timestamp + (now - db_start)`.
-  `auto_timestamp_base` is renamed to `auto_timestamp_base_wall_clock` and used as the DB start wall‑clock reference.
> `auto_timestamp_base_wall_clock` is just the wall‑clock “db_start” anchor used to compute elapsed time when a packet omits a timestamp.
- `set_earliest_timestamp` now persists `time.start_timestamp` and resets the base.
- CLI --start-timestamp and World.run() both flow through `set_earliest_timestamp`.

libs/db/src/lib.rs
```diff
@@ -285,6 +285,8 @@ pub struct DB {
     pub default_stream_time_step: AtomicU64,
     pub last_updated: AtomicCell<Timestamp>,
     pub earliest_timestamp: AtomicCell<Timestamp>,
+    // Wall-clock timestamp at the moment the auto timestamp baseline was set.
+    auto_timestamp_base_wall_clock: AtomicCell<Timestamp>,
```
x46085: rename to "db_start_wall_clock"

prefer no use of "auto_timestamp", it's confusing

libs/db/src/lib.rs
```diff
@@ -411,8 +417,21 @@ impl DB {
...
+    pub fn set_earliest_timestamp(&self, timestamp: Timestamp) -> Result<(), Error> {
+        self.with_state_mut(|state| {
+            state.db_config.set_time_start_timestamp_micros(timestamp.0);
+        });
         self.earliest_timestamp.store(timestamp);
+        self.auto_timestamp_base_wall_clock.store(Timestamp::now());
+        self.save_db_state()
+    }
+
+    pub fn auto_timestamp(&self) -> Timestamp {
```
x46085: rename to "apply_timestamp()"

zxq82lm: "apply_implicit_timestamp()"?

Issue #100, opened 2025-08-25
Update AlephOS to Jetpack 6


x46085:
Jetpack 6 has updates and features that make development with important aspects like cameras more readily accessible. Upgrade to the latest Jetpack / Anduril base upstream dependencies, and test deployment and development on Aleph.

labrnth:
Merged in  #376

PR #417, opened 2026-01-15
fix(aleph): fix b2b DT for peripheral support

labrnth:
fix(aleph): fix device tree source to restore functionality of Jetson SPI and GPIOs on J5 port on expansion board.

See also: https://github.com/elodin-sys/aleph-jetson-orin-baseboard-nvidia-t23x-public-dts/commit/806a0461e17661a7c8bbe4681dc54474d8b36b5e

This fix completes #100 for the purpose of the 16 release.


labrnth:
Tested on base elodin and when used as an imported flake.

PR #422, opened 2026-01-20
fix: tabs persistence.

zxq82lm:
## Content
- Preserve tabs in schematic save so the root `tabs { … }` block is written instead of being flattened.
- Fix load behavior to keep root tabs visible and avoid missing tab bars.
- Persist window `title=` and container `name=` fields during save.
- Avoid the extra "Tabs" tab when loading secondary-window schematics by reusing the root tabs container.


zxq82lm:
> Does this fix “sticky tabs” or no?

@x46085 No. I’m working on “sticky tabs” now.

Issue #439, opened 2026-01-23
Inspector shown too eagerly


shanecelis:
# Problem

When you have multiple tabs, when you click on a tab title, it shows the inspector. This also happens if you click on a tab title whose contents are behind another tab, so effectively you can't select a tab to view without also showing an inspector.

An additional issue is we show an inspector pane even when there is no inspector for that pane. For instance, when I click on a monitor which doesn't have an inspector yet. Perhaps we should not show an inspector if there isn't one available for that tab. (If we will soon have an inspector for everything, then we can ignore this issue.)

# Solution

Only show the inspector on a tab title double-click. N. agrees this would be preferable behavior.

anderspitman:
@shanecelis does this still need work? An empty inspector is still shown for me by default, but I believe the missing inspectors have been added.

shanecelis:
No, this has been fixed. Thanks, @anderspitman.

PR #440, opened 2026-01-23
feat: Right-click on tab title shows context menu to open inspector or rename.

shanecelis:
This fixes issue #439.

### DIFF
--- a/libs/elodin-editor/src/ui/tiles.rs
+++ b/libs/elodin-editor/src/ui/tiles.rs
@@ -1471,18 +1471,18 @@ impl TileState {
     pub fn new(tree_id: Id) -> Self {
         Self {
             tree: egui_tiles::Tree::new_tabs(tree_id, vec![]),
-            tree_actions: smallvec![TreeAction::AddSidebars],
+            tree_actions: smallvec![],
             graphs: HashMap::new(),
             container_titles: HashMap::new(),
-            sidebar_state: SidebarMaskState::default(), // Sidebars start masked (hidden)
+            sidebar_state: SidebarMaskState::default(),
             tree_id,
         }
     }
 
     fn reset_tree(&mut self) {
         self.tree = egui_tiles::Tree::new_tabs(self.tree_id, vec![]);
-        self.tree_actions = smallvec![TreeAction::AddSidebars];
-        self.sidebar_state = SidebarMaskState::default(); // Sidebars start masked (hidden)
+        self.tree_actions = smallvec![];
+        self.sidebar_state = SidebarMaskState::default();
     }
 }
 


libs/elodin-editor/src/ui/tiles.rs
```diff
@@ -1794,6 +1795,8 @@ impl egui_tiles::Behavior<Pane> for TreeBehavior<'_> {
     ) -> egui::Response {
         if button_response.middle_clicked() && !self.read_only {
             self.tree_actions.push(TreeAction::DeleteTab(tile_id));
+        } else if button_response.double_clicked() {
```
anderspitman: Double click is triggering tab rename for me. Are we sure we want this tied to both?

libs/elodin-editor/src/ui/tiles.rs
```diff
@@ -2503,6 +2506,35 @@ impl WidgetSystem for TileLayout<'_, '_> {
                     TreeAction::SelectTile(tile_id) => {
                         tile_state.tree.make_active(|id, _| id == tile_id);
 
+                        if let Some(egui_tiles::Tile::Pane(pane)) =
```
anderspitman, line 2615: This looks identical to the code under the `OpenInspector` branch, unless I'm missing something. Would it make sense to extract into a function? I'm not fanatical about DRY, but do try to lean that direction.

libs/elodin-editor/src/ui/tiles.rs
shanecelis: That's a great point, Anders. I altered this to only show an inspector on a double-click, but we also rename on a double-click. We do need to disambiguate these actions. 

I shudder to suggest a triple click for rename.

Any suggestions? @x46085 @zxq82lm

zxq82lm: I don’t really have a strong opinion yet... Maybe we should ask users what they’d prefer?
Why not right-click on the tile/pane title? That said, I’m not sure that would feel satisfying either.

x46085: Might consider a right click for one of the actions (inspector seems more natural to me)

shanecelis:
How about a context menu? 

https://github.com/user-attachments/assets/0b6e5b32-ff90-4122-8344-1eb7224ec19a

libs/elodin-editor/src/ui/tiles.rs
shanecelis: Resolved to not open inspector on double-click. Instead right-click opens a context menu that shows the options: Show/Hide Inspector, and Rename Tab. Double-click on tab title renames the tab. Click on tab title selects it in the inspector if shown.

libs/elodin-editor/src/ui/tiles.rs
shanecelis: I take your point. They are similar but not identical. If I see a third instance, I'll refactor.

PR #442, opened 2026-01-24
A bunch of UI fixes

shanecelis:
This PR has a number of UI fixes:

* fix: Drag-and-drop of active tab doesn't leave a void.

It ensures a remaining tab is active.

* feat: Add vertical bar between tab titles.

Make the tab titles less contrast. Taking @Teegurks's suggestion.

* hack: Show tabs present but prune them.

  Seems to work well and fixes the dragging leaving a void.

* excise: Remove sidebars from display

  They shouldn't be egui::tiles going forward.
  
* fix: Add margin to frame avoid tabs under the macOS title bar.
* fix: Text labels use theme colors in secondary windows.
* feat: Show tab close on hover only. (Thanks, @Teegurks!)
* ui: In dark theme, make active tabs black on white; make inactive tabs grey on white.
* ui: In dark theme, make active tabs white on black; make inactive tabs grey on black.
* excise: Remove redundant label of monitor in panel.

Screenshot 2026-01-23 at 5 10 04 PM
Screenshot 2026-01-23 at 5 10 10 PM

### DIFF
--- a/libs/elodin-editor/src/ui/mod.rs
+++ b/libs/elodin-editor/src/ui/mod.rs
@@ -377,6 +377,7 @@ impl RootWidgetSystem for MainLayout<'_, '_> {
         #[cfg(not(target_family = "wasm"))]
         world.add_root_widget::<status_bar::StatusBar>("status_bar");
 
+        #[allow(unused_mut)]
         let mut frame = egui::Frame::new();
         #[cfg(target_os = "macos")]
         {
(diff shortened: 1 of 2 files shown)


shanecelis:
I believe what it's showing is correct. I added " and Angles" to visually differentiate. See below.

Screenshot_2026-01-23_at_6 09 10_PM

PR #444, opened 2026-01-27
Add `@elodin.map_seq` to preserve `cond()` behavior.

shanecelis:
# Problem

Elodin map translates jax `cond()` calls into `select()` calls. They both operate like if statements: `cond()` evaluates one of its consequents, but `select()` always evaluates both. 

In general this optimization is good; however, if one of your consequents is very expensive to evaluate, then `select()` forces you to pay the highest cost each time.

# Solution

Provide an escape hatch from this behavior by adding a parameter to `Query.map()` of `use_vmap` which by default is `True`. The original behavior is preserved. But if one passes `use_vmap=False` then your function will be evaluated without vmap and without translating `cond()` into `select()`.

This made a solution that S. was working on feasible and performant. He had a fast path that was used in general, but in an exceptional regime, he had a slow path. Without this feature the solution was too slow to be practical.

### DIFF
--- a/docs/public/content/reference/python-api.md
+++ b/docs/public/content/reference/python-api.md
@@ -616,6 +616,39 @@ def gravity(query: el.Query[el.Force, el.Inertia]) -> el.Query[el.Force]:
     )
 ```
 
+### `@elodin.map_seq`
+
+`@elodin.map_seq` is similar to `@elodin.map`. In fact they will produce the same results numerically but their performance differs.
+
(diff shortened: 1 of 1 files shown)


shanecelis:
Dan, I thank you kindly for your prodding. This design is much clearer and it unlocked some easy wins, namely we can use the decorator `@el.map_seq` as easily as one used `@el.map` before, which I'm sure will be a relief to S.

I added some tests to ensure no semantic difference between `map` or `map_seq` and there are none.

However, one strange bit of behavior is that the test `test_map_vs_map_seq_results_match_batch_size_0` seem to illustrate is that it will panic if a system expects entities with some components and none are present. This seems wrong or I've misapplied it.

PR #448, opened 2026-01-28
Fix UI timeline handle passing sim handle

anderspitman:


shanecelis:
This is going to make a lot of people happy.

There is one thing I would consider trying with respect to this PR: is it possible to have it not surpass the database but remain in a play state? 

The reason I suggest this is it's possible that the simulation runs near real time or slower than real-time, in which case it'll pause as soon as it catches up. I'd prefer if the database is being written to that it merely "buffer", in that it's still in a play state; and if more data is delivered, it will proceed showing that data.

You can alter a sim to force this kind of circumstance by adding some sleeps in the `post_step()`.

anderspitman:
@shanecelis great idea. Just pushed the changes.

shanecelis:
Just tested it with a purposefully slowed down ball example. Here's the main.py:

```python
from sim import SIM_TIME_STEP, system, world

import time
def post_step(tick, ctx):
    time.sleep(0.1)

world().run(system(), sim_time_step=SIM_TIME_STEP, max_ticks=1200, post_step = post_step)
```
[timeline-fix.webm](https://github.com/user-attachments/assets/ad31d762-8ea6-4d4b-b05c-3963aac6322d)

And it works well. This is so _much_ better than before. Thank you, Anders!

PR #453, opened 2026-01-28
Improve command palette state handling

### LINKED ISSUE #435: Command palette remembers its last state erroneously
# Problem

When I hit 'Ctrl-P', the command palette allows me to select the right command. However, if the command I selected requests follow-up input, the command palette shows a white arrow indicating how deep in your command input is. The problem is after that command is finished, and I hit 'Ctrl-P' again, we see it show the prior command input state, and the completions don't quite work.

## Work Around

You can hit 'Backspace' to return to the top-level of the command palette.

## Solution

When 'Ctrl-P' brings up a command palette, it ought to be at the "top level" without any arrows and backspace needed to put it in that mode.

Here is a video showing the behavior.

https://github.com/user-attachments/assets/5f257825-e65c-428a-800f-084eea51d226

anderspitman:
Previously it always remembered nesting. Now, CTRL-P clears the state and starts at the top, while CTRL-SHIFT-P will bring it up with any existing state.

Fixes #435

### DIFF
--- a/libs/elodin-editor/src/lib.rs
+++ b/libs/elodin-editor/src/lib.rs
@@ -238,7 +238,10 @@ impl Plugin for EditorPlugin {
             .add_systems(Startup, spawn_ui_cam)
             .add_systems(PostUpdate, ui::video_stream::set_visibility)
             .add_systems(PostUpdate, set_clear_color)
-            //.add_systems(Update, clamp_current_time)
+            .add_systems(
+                Update,
+                clamp_current_time.before(crate::ui::timeline::timeline_slider::sync_ui_tick),
+            )
             .insert_resource(WireframeConfig {
                 global: false,
                 default_color: Color::WHITE,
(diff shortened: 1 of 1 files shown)


shanecelis:
Looks good. Can you document this in the docs/public/content/reference/command-palette.md?

shanecelis (approved):
Approved once doc changes are in.

anderspitman:
@shanecelis one thought: would this be considering a breaking change? I could definitely see people being used to the way it currently works, and annoyed that they now have to restart from the top every time. They might not check the docs and realize the previous behavior is still available. Maybe we should keep the current behavior, and add `CTRL-SHIFT-P` as the clear state option?

shanecelis:
If we had automated testing scripts, then I would consider it a breaking change, but as it's UI and I often see people hit Ctrl-P and then backspace a bunch, I wouldn't classify it as a breaking change. It's a UI improvement, but I'll try to put that information front and center when we release in case people do like and use the prior behavior.

PR #460, opened 2026-02-02
Fix python sim CTRL-C

anderspitman:


anderspitman:
@shanecelis @zxq82lm this isn't quite ready for review. Still need to test on windows. But can one of you do a quick sanity check that it works on macOS when you get a chance?

zxq82lm:
@anderspitman I just tested the branch on macOS (drone example), and Ctrl+C works fine.

anderspitman:
@zxq82lm thanks!

anderspitman:
This is ready for review. One potential issue. I stuck with the `signal-hook` crate, but changed the way we're using it. There's a note on the crate docs that this API isn't compatible with windows. However, I believe we're only using `nox-py` in WSL on windows. At least I've never successfully built it for native, and our CI/CD doesn't. Also, I think Windows might force close on CTRL-C anyway, sidestepping this issue. I left a comment in the code. Basically I think it's fine as is, and if we add native windows builds in the future and there's a problem with CTRL-C, we'll know where to look.

Alternatively I could switch to the `ctrlc` crate now, which is smaller in scope and advertises windows support, but we can't really test whether there's a difference in behavior so I'm not sure it's worth it.

anderspitman:
@shanecelis that's really interesting. I'm inclined to try and understand why this isn't needed on your setup. Are you running them from the editor like this:

```
elodin run examples/ball/main.py
```

or directly with python:

```
python3 examples/ball/main.py
```

We can also just move on if it's not worth the time.

shanecelis:
I am using `elodin editor examples/ball/main.py`. Let me try with the other two and on macOS.

shanecelis:
When I run with `python3` on main on Linux I can get a ignored Ctrl-C that this branch does respect, so that's a confirmation of this being better.

shanecelis (changes requested):
~~macOS Ctrl-C not working.~~

Correction: macOS Ctrl-C on my erroneous branch was not working, but it is working on this branch. Further confirmation to accept it to main.

PR #461, opened 2026-02-02
Animate joint on GLB file via EQL.

### LINKED ISSUE #452: Drive articulation in 3d model via EQL queries.
# Problem

Given a rigged 3d-model, we don't have a means of animating it.

# Solution

There ought to be a way to drive 3d model animation from an EQL expression. 

For instance, suppose I had a rigged plane model whose flaps could be controlled by specifying the angles of each. I'd want to specify something like this in KDL:

```kdl
object_3d ball.world_pos {
    animate joint="left-flap" value="plane.flap[0]"
    animate joint="right-flap" value="plane.flap[1]"
        color 255 128 0
    }
}
```

shanecelis:
Fixes #452.

### DIFF
--- a/libs/impeller2/kdl/src/ser.rs
+++ b/libs/impeller2/kdl/src/ser.rs
@@ -450,34 +450,29 @@ fn serialize_object_3d_mesh(mesh: &Object3DMesh) -> (KdlNode, Vec<KdlNode>) {
             let node = match mesh {
                 Mesh::Sphere { radius } => {
                     let mut node = KdlNode::new("sphere");
-                    node.entries_mut()
-                        .push(KdlEntry::new_prop("radius", *radius as f64));
+                    push_rounded_float_prop(&mut node, "radius", *radius as f64);
                     serialize_material_to_node(&mut node, material);
                     node
                 }
                 Mesh::Box { x, y, z } => {
                     let mut node = KdlNode::new("box");
-                    node.entries_mut().push(KdlEntry::new_prop("x", *x as f64));
-                    node.entries_mut().push(KdlEntry::new_prop("y", *y as f64));
-                    node.entries_mut().push(KdlEntry::new_prop("z", *z as f64));
+                    push_rounded_float_prop(&mut node, "x", *x as f64);
+                    push_rounded_float_prop(&mut node, "y", *y as f64);
+                    push_rounded_float_prop(&mut node, "z", *z as f64);
                     serialize_material_to_node(&mut node, material);
                     node
                 }
                 Mesh::Cylinder { radius, height } => {
                     let mut node = KdlNode::new("cylinder");
-                    node.entries_mut()
-                        .push(KdlEntry::new_prop("radius", *radius as f64));
-                    node.entries_mut()
-                        .push(KdlEntry::new_prop("height", *height as f64));
+                    push_rounded_float_prop(&mut node, "radius", *radius as f64);
+                    push_rounded_float_prop(&mut node, "height", *height as f64);
                     serialize_material_to_node(&mut node, material);
                     node
                 }
                 Mesh::Plane { width, depth } => {
                     let mut node = KdlNode::new("plane");
-                    node.entries_mut()
-                        .push(KdlEntry::new_prop("width", *width as f64));
-                    node.entries_mut()
-                        .push(KdlEntry::new_prop("depth", *depth as f64));
+                    push_rounded_float_prop(&mut node, "width", *width as f64);
+                    push_rounded_float_prop(&mut node, "depth", *depth as f64);
                     serialize_material_to_node(&mut node, material);
                     node
                 }


shanecelis:
Try running this new example to confirm its behavior:

```sh
cargo elodin editor examples/animated-rocket/main.py
```

anderspitman:
@shanecelis I'm seeing some flickering in the charts on this branch. It happens on the old rocket as well:


https://github.com/user-attachments/assets/693eb909-c0c5-4d27-a689-d9253da28d79

shanecelis:
I saw that too. Good capture of it. I say file an issue using that capture. I'm glad we can reproduce it.

anderspitman:
I think it's specific to this branch, or are you saying you've observed this on main?

shanecelis:
Oh, I see. I misread. I thought you said you'd seen it elsewhere. I'll look into it.

anderspitman:
Actually I just realized I was very unclear, sorry about that. What I meant by "old rocket" is `examples/rocket`, but only on this branch, not main.

shanecelis:
Re-basing fixed the graph flickering issue. But I am going to try to alter the crazyflie example to provide an example of the animation.

shanecelis:
https://github.com/user-attachments/assets/f0fbf881-7e7f-4361-ac0b-cfe499f4e053

shanecelis:
I removed the animated-rocket example and enhanced the crazyflie-edu example.

https://github.com/user-attachments/assets/522ec3d3-8b9e-4362-84d6-ebcf732a9a52

shanecelis:
I changed the KDL `rotation_vector` to use degrees instead of radians because that's used by `glb`'s `rotate` vector as well.

I moved the crazyflie model over by -0.01 on the x-axis to center it, which was done before in the KDL. It worked fine, but when `rotate` was applied to the KDL `glb path="..." rotate="(0,90,0)"` it would rotate around the model's center. I did this to ensure that the KDL can control the orientation of its vehicle now that the animated fins are going to be relevant to analysis.

anderspitman:
Here's another comment from Codex:

```
joint animation expressions are not refreshed after initial attach in common edit flows.
      - attach_joint_animations only runs when on_scene_ready is true (libs/elodin-editor/src/object_3d.rs:1156), and
        also skips entities that already have the component (libs/elodin-editor/src/object_3d.rs:773).
      - Impact: if rotation_vector/joint mapping changes after scene load, existing JointAnimationComponent keeps stale
        compiled expressions.
   ```

**Comment by shanecelis** [comment]
Nice finds, Anders. Will fix.

**Comment by shanecelis** [comment]
The round tripping in KDL is fixed.

I've tested the refresh, and there are two things. If I reload the KDL, it works. I think internally everything is scrapped, but it works on a reload.

If I edit the KDL, I don't see any automatic refresh. Perhaps that's a regression.

**Comment by anderspitman** [comment]
@shanecelis I don't see any code changes. If you're ok with the way it's currently working, I can approve.

**Comment by shanecelis** [comment]
You're right, @anderspitman. I'm certain I have commits that add the test you suggested and a bug fix. Let me find it.

**Comment by anderspitman** [comment]
@shanecelis most of the new stuff since my last review seems to be the up/direction bits. Do you have an example of how that works so I can test it? I may have just missed it...

**Comment by shanecelis** [comment]
Well, here you can see the ball upside down or fly behind the wheel of the drone to confirm they work.

```diff
index 6a95ac84..3d7b8d38 100644
--- a/examples/ball/sim.py
+++ b/examples/ball/sim.py
@@ -26,7 +26,7 @@ def world(seed: int = 0) -> el.World:
         """
         hsplit {
             tabs {
-                viewport name=Viewport pos="(0,0,0,0, 8,2,4)" look_at="(0,0,0,0, 0,0,3)" hdr=#true show_grid=#true active=#true
+                viewport up="(0,0,-1)" name=Viewport pos="(0,0,0,0, 8,2,4)" look_at="(0,0,0,0, 0,0,3)" hdr=#true show_grid=#true active=#true
                 inspector
             }
         }
diff --git a/examples/drone/sim.py b/examples/drone/sim.py
index 834cd2ee..23f3d4d9 100644
--- a/examples/drone/sim.py
+++ b/examples/drone/sim.py
@@ -138,7 +138,7 @@ def world() -> tuple[el.World, el.EntityId]:
 
         tabs {
             hsplit name = "Viewport" {
-                viewport name=Viewport pos="drone.world_pos + (0,0,0,0, 2,2,2)" look_at="drone.world_pos" show_grid=#true active=#true
+                viewport up="drone.world_pos.direction(0,0,1)" name=Viewport pos="drone.world_pos.translate(-2,0,0)" look_at="drone.world_pos" show_grid=#true active=#true
                 vsplit share=0.4 {
                     graph "drone.angle_desired" name="angle_desired"
                     graph "drone.world_pos.q0, drone.world_pos.q1, drone.world_pos.q2, drone.world_pos.q3, drone.attitude_target" name="World Pos"
```

PR #478, opened 2026-02-12
Add `elodin editor --kdl my-kdl.kdl` CLI option.

shanecelis:
In testing I've wanted to be able to set my KDL from the CLI. It's become something of a rote routine to hit Ctrl-P -> Load Schematic -> pick your schema when working with an existing database. And it's definitely what I see happen in practice. However, we can avoid that routine with this `--kdl PATH` option.


zxq82lm (approved):
May be add a brief explanation in the Elodin CLI documentation?
https://docs.elodin.systems/reference/elodin-cli/

PR #479, opened 2026-02-12
Implement KDL float rounding

### LINKED ISSUE #220: Add float rounding in KDL serialization
## Summary
KDL output currently mirrors raw floats, so a clean value like sphere radius=0.2 ends up serialized as 0.20000000298023224. We should round floats during serialization so saved schematics stay readable and stable.

Ball Example is saved in:
```
tabs {
    viewport name=Viewport pos="(0,0,0,0, 8,2,4)" look_at="(0,0,0,0, 0,0,3)" show_grid=#true
}
object_3d ball.world_pos {
        color 255 128 0
    }
}
    color 255 255 255
}
```

## Notes
- Applies to all serialize_* helpers that emit floats (schematics, object definitions, etc.).
- Consider a shared helper (`round_float(value, precision)`) used when pushing `KdlEntry`s.
- Pick a default precision (6 decimals?) and document it.

## Acceptance
- Saving a schematic with floats like 0.2 keeps them short (no long tail of digits).

anderspitman:
For values like 0.2, it was emitting things like
0.20000000298023224. Now rounds with precision 6 and emits 0.2.

Fixes #220

### DIFF
--- a/libs/elodin-editor/src/plugins/gizmos.rs
+++ b/libs/elodin-editor/src/plugins/gizmos.rs
@@ -1,5 +1,6 @@
 use bevy::camera::RenderTarget;
 use bevy::camera::visibility::RenderLayers;
+use bevy::picking::prelude::Pickable;
 use bevy::ui::{Node, PositionType, UiTargetCamera, Val, ZIndex};
 use bevy::window::WindowRef;
 use bevy::{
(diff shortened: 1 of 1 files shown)


anderspitman:
Hm that's a weird check fail

anderspitman:
Tests are working for me locally. Is there a way to just rerun it?

x46085:
> Tests are working for me locally. Is there a way to just rerun it?

Yea this test is flakey, something like 1/8 runs: "Rust: maitake v0.1.0"

I've tried to fix this a few times, still lingers.

I've queued it for running again.

PR #486, opened 2026-02-13
Feat: Display replay speed.

zxq82lm:
## Content
- This branch shows the current playback speed in the timeline controls (SPEED, e.g. [0.5x]).

### DIFF
--- a/libs/db/src/lib.rs
+++ b/libs/db/src/lib.rs
@@ -332,6 +332,13 @@ pub struct DB {
     pub earliest_timestamp: AtomicCell<Timestamp>,
     // Wall-clock timestamp at the moment the DB start anchor was set.
     db_start_wall_clock: AtomicCell<Timestamp>,
+    /// When true, last_updated advances with playback position instead of
+    /// reflecting the actual data extent. This makes recorded data look like
+    /// live telemetry to connected editors.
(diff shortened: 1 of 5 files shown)


shanecelis:
image

x46085:
Nice! I disabled the fetching the speed from the database here:
https://github.com/elodin-sys/elodin/pull/485

Can you update this to work with it?


@x46085 This one is done.

x46085:
What happens if you use the world.run API and pass the playback speed in to some value that isn’t one of the drop downs?

zxq82lm:
@x46085 I'm on these two points:
- Update regarding #485.
- Usage with world.run API?

zxq82lm:
> What happens if you use the world.run API and pass the playback speed in to some value that isn’t one of the drop downs?

@x46085:
If you pass a playback speed via world.run that isn’t one of the dropdown values, the editor now ignores it on connect and resets playback to 1.0x (1/60s timestep). In practice, only speeds explicitly set through the editor controls are applied/displayed.

Should I change this behavior?

x46085:
> @x46085: If you pass a playback speed via world.run that isn’t one of the …
> Should I change this behavior?

No that works for me

x46085 (approved):
Tested locally, looks / works great.

PR #492, opened 2026-02-15
feat: Add zoom in button on hold alt.

shanecelis:
Zoom out is so useful it deserves its reciprocal zoom in. Currently activated by holding the alt key.

https://github.com/user-attachments/assets/8ddf55f1-83cf-4f85-beca-55a786e8c79a


zxq82lm (approved):
Super idea!
Just tested, it works so well.
(On macOS, the button to maintain to ZoomIn is "Option".)

x46085:
Nice!

Since you’re in here, do you see how would could add to the KDL a show/hide option for a given viewport? Actually we do that for the grid already, I suppose that pattern would work. I need to turn off the cube for the RC Jet chase cam, and a few other constrained views.  I wonder if there is a window size that we should auto hide it when it’s too big for the space practically

zxq82lm:
> Nice!
> 
> Since you’re in here, do you see how would could add to the KDL a show/hide option for a given viewport? Actually we do that for the grid already, I suppose that pattern would work. I need to turn off the cube for the RC Jet chase cam, and a few other constrained views. I wonder if there is a window size that we should auto hide it when it’s too big for the space practically

We already have `show_view_cube` (default true) on viewport-kdl. Should we improve it?

PR #493, opened 2026-02-15
fix: Camera reset behavior. [abandoned]

zxq82lm:
## Why this PR

The camera reset (from the ViewCube and the Command Palette) was setting `Transform::IDENTITY`, which effectively teleported the camera back to the world origin.

## Expected behavior
A camera reset should:
- reset the direction/orientation,
- stop any ongoing movement,
- without changing the camera’s current position.


x46085:
The reset function is meant to return it to whatever is specified in the KDL originally, not just fix orientation. For example if I’ve zoomed and panned the camera a bit to look around a scene, and I want to go back to the KDL original setting, click the button. Does it not do this already?

@x46085 Yes, that’s the current behavior. Indeed, it should remain as it is.

PR #508, opened 2026-02-19
Feat: Frustum displaying from KDL.

zxq82lm:
## Viewport parameters
- `create_frustum` (bool): create/publish this viewport camera frustum.
- `show_frustums` (bool): render frustums created by other viewports on this viewport.
- `near`/`far` (optional): override camera clipping planes. Defaults are `near=0.05`, `far=5.0`.
- `aspect` (optional): override camera aspect ratio.
  - if omitted: aspect is derived from the real viewport size (`width / height`)
  - if set: camera projection (and frustum shape) uses this fixed ratio
- `frustums_color` (optional): named color or tuple string like `"(255,255,0,200)"`.
- `frustums_thickness` (optional): edge radius in world units.

## KDL usage
```kdl
tabs {
    viewport name=ViewportA \
        pos="(0,0,0,0, 8,2,4)" \
        look_at="(0,0,0,0, 0,0,0)" \
        create_frustum=#true \
        show_frustums=#true \
        frustums_color="yalk" \
        frustums_thickness=0.008 \
        near=0.05 \
        far=42.0 \
        aspect=1.7778 \
        active=#true

    viewport name=ViewportB \
        pos="(0,0,0,0, 2,2,2)" \
        look_at="(0,0,0,0, 0,0,0)" \
        show_frustums=#true \
        active=#true
}
```

## Viewport Inspector
- `fov`, `near`, `far`, and `show_frustums` are editable.
- `aspect` can be managed as Auto (from viewport size) or Fixed (user-defined ratio).
- `create_frustum` is a toggle action on the same button: `CREATE` when absent, `DELETE` when already created.
- When `create_frustum` is enabled: `frustums_color` and `frustums_thickness` are editable.
   - When a frustum is created, its `frustums_color` is shown as a small dot next to the viewport title.

## Fix
- Inspector header now shows the selected viewport panel title (same behavior as Graph). 15f347fce10612a36ed53e4f94142e46a6f0c53c

---
---

https://github.com/user-attachments/assets/ea6eaf6e-912b-4a5d-8659-be80b911a810

---
---

```kdl
vsplit share=0.5 {
 viewport name=Viewport pos="drone.world_pos + (0,0,0,0, 2,2,2)" look_at="drone.world_pos" show_grid=#true create_frustum=#true show_frustums=#true active=#true
 viewport name=RearViewport pos="drone.world_pos.translate_x(-3.0).translate_z(1.0)" look_at="drone.world_pos.translate_x(1.5)" show_grid=#true create_frustum=#true show_frustums=#true
}
```

image

### DIFF
--- a/libs/elodin-editor/src/ui/schematic/mod.rs
+++ b/libs/elodin-editor/src/ui/schematic/mod.rs
@@ -142,12 +142,18 @@ impl SchematicParam<'_, '_> {
                             .ok()
                             .and_then(|projection| match projection {
                                 Projection::Perspective(perspective) => {
-                                    let near = if (perspective.near - tiles::DEFAULT_VIEWPORT_NEAR).abs() > f32::EPSILON {
+                                    let near = if (perspective.near - tiles::DEFAULT_VIEWPORT_NEAR)
+                                        .abs()
+                                        > f32::EPSILON
+                                    {
                                         Some(perspective.near)
                                     } else {
                                         None
                                     };
-                                    let far = if (perspective.far - tiles::DEFAULT_VIEWPORT_FAR).abs() > f32::EPSILON {
+                                    let far = if (perspective.far - tiles::DEFAULT_VIEWPORT_FAR)
+                                        .abs()
+                                        > f32::EPSILON
+                                    {
                                         Some(perspective.far)
                                     } else {
                                         None


x46085:
Oh, this is quite a bit different from what I was expecting. 

For a camera view frustum, I would want to create a viewport, define it's FOV and position in the world like normal, and toggle on it's frustum for display (create_frustrum=true). Then other viewports would see it, and for disabling it in a specific viewport use show_frustums=false. I wasn't expecting to involve 3d_object at all.



For this API, how would I position the frustum to a specific position relative to the 3d_object? For the drone case, I'd want to put it at the nose of the drone, where the camera is. I don't see how to do that here. Or is the frustrum meant to be it's own 3d object definition, and not with the drone?

zxq82lm:
> Oh, this is quite a bit different from what I was expecting.
> 
> For a camera view frustum, I would want to create a viewport, define it's FOV and position in the world like normal, and toggle on it's frustum for display (create_frustrum=true). Then other viewports would see it, and for disabling it in a specific viewport use show_frustums=false. I wasn't expecting to involve 3d_object at all.
> 
> For this API, how would I position the frustum to a specific position relative to the 3d_object? For the drone case, I'd want to put it at the nose of the drone, where the camera is. I don't see how to do that here. Or is the frustrum meant to be it's own 3d object definition, and not with the drone?

----
- Next test: I’m going to make the following change -> treat the Viewport’s fov as the reference (no more fov defined on the 3D Frustum object).
-  @x46085: Maybe I should open an RFD on the Frustum topic?

shanecelis:
Benoît brought up a use case I hadn't considered, which is why I encouraged him to make frustums drawable absent a camera: Suppose we have a video_stream from an external camera. That's a camera we don't have a projection matrix for. So it seemed to me that we should be able to draw frustums without cameras. 

However, we should still also be able to easily draw a frustum, and exactly that camera's frustum, for any viewport with "show_frustum=#true".

x46085:
> Benoît brought up a use case I hadn't considered, which is why I encouraged him to make frustums drawable absent a camera: Suppose we have a video_stream from an external camera. That's a camera we don't have a projection matrix for. So it seemed to me that we should be able to draw frustums without cameras.
> 

That makes sense, and I'm open to building for both as separate APIs. I just don't want to have to create a 3d_object for my viewport camera frustrum.

x46085:
> * `show_frustum=#true` on a viewport, two cases:
>   
>   * case (1): If `near=` and `far=` are also defined on this viewport, we display that viewport’s frustum.
>   * case (2): If `near=` or `far=` are not defined, then **if is exists** a 3D frustum object defined in the KDL, we display that 3D frustum object.

No this seems like we're going down a bad path.

Just stick with the API you have (3d_object based) and forget about trying to put frustums natively on viewports.

However one API change for "show_frustum=#true", please make this `show_frustum` -> `show_frustums`, since this is a renderlayer toggle for any frustum objects in the scene, not just one.

PR #514, opened 2026-02-22
Fix timestamp from real world

x46085:
> [!NOTE]
> **Medium Risk**
> Touches core time-range bookkeeping (`earliest_timestamp`/`last_updated`) during ingest and `DB::open()`, which can affect timeline behavior across all datasets, but the change is guarded and covered by a new regression test.
> 
> **Overview**
> Fixes time-range initialization when databases contain *dense data starting at timestamp 0*, avoiding a degenerate `(earliest >= last_updated)` range that made the editor timeline explode.
> 
> `DB::open()` now detects this degenerate case and falls back to scanning component start timestamps (excluding timestamp-source components) to recover a correct `earliest_timestamp`. Runtime timestamp tracking is also updated in `DBSink` and follower `TimeSeries` ingest to *not* update `earliest_timestamp`/`last_updated` from timestamp-source components, and the editor now defaults to a small `0..1_000_000` range when it has no data and the selection is still uninitialized. Adds a regression test covering the epoch-zero scenario.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2a6a1bb41b9d98d3a63648ddd54a0d9454f546b0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/libs/db/src/follow.rs
+++ b/libs/db/src/follow.rs
@@ -335,9 +335,18 @@ async fn run_follower_inner(config: &FollowConfig, db: &Arc<DB>) -> Result<(), E
                             .push_buf(timestamp, &data_buf[start..end]);
                     }
                     if timestamps.len() > skip {
-                        db.last_updated.update_max(*timestamps.last().unwrap());
-                        if let Some(&first_valid) = timestamps[skip..].iter().find(|ts| ts.0 > 0) {
-                            db.earliest_timestamp.update_min(first_valid);
+                        let is_ts_source = state
+                            .component_metadata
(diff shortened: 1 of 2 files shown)


zxq82lm (approved):
LGTM. 

The DB fallback correctly recovers epoch-zero datasets without affecting normal positive-timestamp paths. 

**Minor note:** >= also triggers on single-timestamp DBs (harmless but logs a spurious warning); could tighten to > if the noise bothers you.

PR #515, opened 2026-02-23
Feat/tracy

anderspitman:


apps/elodin/src/cli/mod.rs
```diff
@@ -41,15 +41,32 @@ impl Cli {
...
             .with_target(false)
-            .with_env_filter(filter)
-            .with_timer(ChronoLocal::new("%Y-%m-%d %H:%M:%S%.3f".to_string()))
+            .with_timer(ChronoLocal::new("%Y-%m-%d %H:%M:%S%.3f".to_string()));
+
+        #[cfg(feature = "tracy")]
+        let init_res = tracing_subscriber::registry()
+            .with(filter)
+            .with(fmt_layer)
+            .with(tracing_tracy::TracyLayer::default())
```
anderspitman, line 56: This adds tracy support to the tracing subscriber we were already creating.

apps/elodin/Cargo.toml
```diff
@@ -10,6 +10,7 @@ authors = ["sascha@elodin.systems", "akhil@elodin.systems", "van@elodin.systems"
 cargo-clippy = []
 inspector = ["elodin-editor/inspector"]
 debug = ["elodin-editor/debug"]
+tracy = ["elodin-editor/tracy", "bevy/trace_tracy", "dep:tracing-tracy"]
```
anderspitman, line 13: Adds `--features tracy` to elodin

libs/elodin-editor/src/ui/mod.rs
```diff
@@ -546,6 +547,7 @@ impl RootWidgetSystem for ViewportOverlay<'_, '_> {
     }
 }
 
+#[instrument]
```
anderspitman, line 551: Example trace for our `render_layout` system

nix/pkgs/tracy.nix
```diff
@@ -0,0 +1,173 @@
+{
```
anderspitman, line 1: This file is Codex's recreation of https://github.com/SanderVocke/tracy-nix

Unfortunately mainline tracy is stuck on an old version because tracy moved to cmake CPM for package management, which doesn't play well with nix. That's why the hacks to download the packages manually in this file.

nix/shell.nix
```diff
@@ -226,9 +224,36 @@ in {
...
+  # Unified shell that combines all development environments
+  elodin = mkShell (shellAttrs // linuxShellAttrs);
+
+  # Profiling shell adds Tracy GUI without duplicating the base shell definition
+  elodin-profiling = mkShell (
+    (shellAttrs
+      // {
+        name = "elo-profiling-shell";
+        buildInputs = shellAttrs.buildInputs ++ [tracy];
+        shellHook = lib.replaceStrings
```
anderspitman: Append a few additional instructions to our `shellHook` for profiling

anderspitman:
`alejandra` nix formatter made a bunch of changes for some reason. I recommend reviewing with whitespace ignored.

anderspitman:
Just pushed an important commit. The previous shellHook instructions told you to build in debug mode, which results in terrible performance and is obviously useless for profiling.

anderspitman:
Also I have no way to test this on macOS. I'd estimate it's pretty likely to not work there currently, though I don't think there are any hard blockers to make it work.

shanecelis:
Tracy was relatively easy to get working on macOS, so I don't mind it being left out. Our most important platform is Linux.

x46085:
> Tracy was relatively easy to get working on macOS, so I don't mind it being left out. Our most important platform is Linux.

hard disagree!

Everything we do in the Elodin codebase needs to stay multi-platform. If we want to drop support just down to Linux, that can be a decision, but we'd do it repo-wide.

shanecelis:
I stand corrected. Thanks for chiming in, Dan.

PR #521, opened 2026-02-23
Fix: Backspace behavior in command palette on Linux.

zxq82lm:
## Summary

On Linux (X11/XWayland via nix develop), pressing Backspace in the command palette search field had no effect — users couldn't delete typed characters and had to Escape and re-open.

Two separate issues were at play:

1. **IME capturing key events (root cause):** bevy_egui's `process_ime_system` calls `set_ime_allowed(true)` whenever a TextEdit gains focus. On Linux, this causes the display server to route Backspace and arrow key events through XIM instead of delivering them as regular `KeyboardInput` events. This is a known egui bug ([emilk/egui#5008](https://github.com/emilk/egui/issues/5008)) that was fixed in egui-winit ([emilk/egui#5188](https://github.com/emilk/egui/pull/5188), [emilk/egui#5198](https://github.com/emilk/egui/pull/5198)), but bevy_egui has its own input pipeline and still triggers the problematic path. Fixed by setting `enable_ime = false` on `EguiContextSettings` on Linux.

2. **Unnecessary page-pop on Backspace:** The old code tried to pop the command palette page stack on every Backspace press with an empty filter, even on the top-level page (single entry). This caused a pointless pop-and-recreate cycle. Fixed by adding a `page_stack.len() > 1` guard and switching from read-only `key_pressed()` to `consume_key()` for proper event consumption.

## Test plan

- [x] Open command palette on Linux, type text, verify Backspace deletes characters normally
- [x] Verify arrow keys work in the command palette search field on Linux
- [x] Navigate into a sub-page, clear the filter, press Backspace — should return to the previous page
- [x] On the top-level page with empty filter, press Backspace — should be a no-op
- [x] Verify no regression on macOS

### DIFF
--- a/libs/elodin-editor/src/lib.rs
+++ b/libs/elodin-editor/src/lib.rs
@@ -349,6 +349,13 @@ fn setup_egui_global_system(mut egui_global_settings: ResMut<EguiGlobalSettings>
 fn setup_egui_context(mut contexts: Query<&mut EguiContextSettings>) {
     for mut context in &mut contexts {
         context.capture_pointer_input = false;
+        // Workaround for https://github.com/emilk/egui/issues/5008
+        // On Linux, IME activation via set_ime_allowed(true) causes the compositor to
+        // capture Backspace/arrow key events, preventing them from reaching TextEdit.
+        #[cfg(target_os = "linux")]
+        {
+            context.enable_ime = false;
+        }
     }
 }
 


zxq82lm:
- Linux-tested successfully on b99eb17b1606df2cd01bf4f87ccd130a9a10491
- No regression on macOS

PR #522, opened 2026-02-23
Enhance ellipsoids for covariance matrix visualization

shanecelis:
# Problem

We want to represent error covariance matrices for 3d positions of objects. These can be represented with an ellipsoid to show where an object could be within a certain confidence interval.

# Solution

Accept a Cholesky decomposition L of the covariance matrix P, P = LL^T. Allow the user to select the confidence interval. (N. suggested 70% with 3 degrees of freedom as default.)


shanecelis:
To exercise this commit, you can refer to the schematic.md document for the new "ellipsoid" attributes, but here is a working example for rocket/main.py one can use:

```kdl
    object_3d "(0,0,0,1, rocket.world_pos[4],rocket.world_pos[5],rocket.world_pos[6])" {
        // ellipsoid scale="(3,2,1)" {
        ellipsoid error_covariance_cholesky="(3, 0, 2, 0, 0, 1)" show_grid=#true error_confidence_interval="70" {
            color 255 200 100 100
            grid_color 255 255 255 255
        }
    }
```

This sets up an ellipsoid at the rocket's position (but does not adopt its orientation). The error_covariance_cholesky describes a lower triangular matrix that looks like this:
```
| 3 _ _ |
| 0 2 _ |
| 0 0 1 |
```
The blanks `_` will be zero, left out for clarity of representation.

The above matrix should be proportionate to a "(3, 2, 1)" scale (this did require a bug fix to make it also ENU). If one chooses a confidence interval of "20" then it should be equivalent.

PR #526, opened 2026-02-26
Feat: Add frustum/ellipsoid intersection features.

zxq82lm:
## Summary

Adds frustum/ellipsoid intersection support in the editor and exposes it in the viewport inspector.

## Changes

- **Viewport controls**: Independent toggles for Frustums, Coverage, Proj. 2D, and Proj. Color.
- **Coverage component**: Frustum coverage stored as a component; live % coverage monitor in the target viewport.
- **Auto-enable**: When `show_frustum=#true` or `show_frustums=#true` is set in the schematic, Coverage and Proj. 2D are enabled automatically.
- **Visibility**: Intersection options shown only when frustum rendering is enabled and an ellipsoid is detected.

## examples/ellipsoid

New `examples/ellipsoid` example: two viewports (one with `create_frustum=#true`, one with `show_frustums=#true`), demonstrating the frustum intersection overlay.

image

https://github.com/user-attachments/assets/326841b1-83df-40a6-972d-2e803e4faf61

### DIFF
--- a/libs/elodin-editor/src/plugins/frustum_intersection/coverage_db.rs
+++ b/libs/elodin-editor/src/plugins/frustum_intersection/coverage_db.rs
@@ -47,10 +47,7 @@ fn set_coverage_value(
 
 /// Persist coverage ratios to impeller2 components. Spawns new entities for unseen ellipsoids,
 /// updates existing ones, and resets coverage to 0.0 for ellipsoids no longer in any frustum.
-pub(super) fn write_coverage_to_db(
-    mut params: CoverageDbParams<'_, '_>,
-    mut commands: Commands,
-) {
+pub(super) fn write_coverage_to_db(mut params: CoverageDbParams<'_, '_>, mut commands: Commands) {
     let mut ratios_by_ellipsoid: HashMap<Entity, f32> = HashMap::new();
     for ratio in params.ratios.0.iter() {
         let entry = ratios_by_ellipsoid.entry(ratio.ellipsoid).or_insert(0.0);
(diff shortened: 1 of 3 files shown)


x46085:
All new editor features need to be represented in the examples somewhere. Our examples support our documentation and we always want to have code in the repo that expresses features that exist.

zxq82lm:
> All new editor features need to be represented in the examples somewhere. Our examples support our documentation and we always want to have code in the repo that expresses features that exist.

---

Starting from the video configuration (one ellipsoid, two viewports) with `create_frustum=#true` on the first viewport and `show_frustums=#true` on the second viewport. Would this be a relevant new example?

---

Other point: the two Intersects features do not yet exist in the KDL. Should I add `intersect_coverage=#true/false` and `intersect_2D=#true/false`, with the constraint that `#true` would only make sense if `show_frustums=#true` && an ellipsoid has been detected?

What do you think about these two points? @anderspitman @shanecelis @x46085

x46085:
> What do you think about these two points? @anderspitman @shanecelis @x46085

I would finish the KDL features needed, and create a new example that expresses everything clearly. This feature is probably too complex to stick into an existing example easily.

zxq82lm:
> > All new editor features need to be represented in the examples somewhere. Our examples support our documentation and we always want to have code in the repo that expresses features that exist.
> 
> Starting from the video configuration (one ellipsoid, two viewports) with `create_frustum=#true` on the first viewport and `show_frustums=#true` on the second viewport. Would this be a relevant new example?
> 
> Other point: the two Intersects features do not yet exist in the KDL. Should I add `intersect_coverage=#true/false` and `intersect_2D=#true/false`, with the constraint that `#true` would only make sense if `show_frustums=#true` && an ellipsoid has been detected?
> 
> What do you think about these two points? @anderspitman @shanecelis @x46085

Another existing option: `show_frustums=#true` would enable both `intersect_ellipsoid` features by default (each could be disabled individually in the inspector).

x46085:
> Another existing option: `show_frustums=#true` would enable both `intersect_ellipsoid` features by default (each could be disabled individually in the inspector).

Yea that would be ideal

PR #532, opened 2026-02-27
Fix(Cube-Viewer): ZoomIn/ZoomOut with floating origin.

zxq82lm:
## Problem
ZoomIn/ZoomOut buttons on the ViewCube did not work correctly with distant objects (e.g. rocket) — the camera lost the target after zooming.

## Solution
- Convert `orbit_target` to floating-origin space before distance calculations (it was in absolute world coordinates while the camera uses floating-origin space)
- Add `Without<>` filters to resolve B0001 query conflict between FloatingOrigin and ViewCubeTargetCamera queries


x46085:
Always best to post a little before / after video capture of the problem and the fix in place, makes this much faster to review and accept without delay.

PR #533, opened 2026-03-02
Nix develop quick

anderspitman:
This adds `nix develop .#quick`, which starts a nix shell without compiling `elodin-py` and `elodinsink`, resulting in much faster startup. This is useful if you're working on our nix files themselves, or if you've made a change that you know doesn't touch the python side, letting you iterate faster.


x46085 (changes requested):
Shane can tell you I famously dislike nix develop shell variants. There actually used to be 3 in here and I did the hard work to kill them. I understand your pain is valid, but rather than work around it, why don't we solve the root problem? Do you have access to the remote builders? Could that be why we don't see this issue but you do?

anderspitman:
@x46085 still getting familiar with nix so it's very likely there is a more idiomatic solution. I don't know anything about the remote builders. What do they do?

Just so I'm clear on the problem I'm trying to solve, any time anything in our repo changes, `elodin-py` and `elodinsink` are rebuilding for me when I run `nix develop`. Huge time sink.

Another possible reason you're not hitting this is because `elodin-py` only builds on Linux.

PR #539, opened 2026-03-04
Fix: Cube-viewer no longer attached to the rendered viewport but to the UI pane.

zxq82lm:
## Problem

When changing the camera aspect= of the viewport, the cube viewer would move.

##  Solution

The cube viewer is no longer attached to the rendered viewport but to the UI pane. This also allows the cube viewer to be positioned anywhere on the viewport according to the user’s preference.

https://github.com/user-attachments/assets/2f9f5b7a-45f2-4868-b81c-9b6c44d22ddd

---

Includes also this fix: Fixed an overly noisy warning that was triggered in valid ellipsoid setups. e45d86cd5c5eed28d5ede9ac7b3600a016f39471

### DIFF
--- a/libs/elodin-editor/src/object_3d.rs
+++ b/libs/elodin-editor/src/object_3d.rs
@@ -746,6 +746,7 @@ pub fn update_object_3d_system(
     )>,
     mut transforms: Query<&mut Transform>,
     mut mat3_params: Query<&mut Mat3Params>,
+    mesh_child_markers: Query<(), With<Object3DMeshChild>>,
     entity_map: Res<EntityMap>,
     component_value_maps: Query<&'static ComponentValue>,
 ) {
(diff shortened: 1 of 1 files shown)


shanecelis:
I like that this is part of UI now. However, dragging it around the screen is less useful from my perspective than being able to grab the cube or near the cube and drag to freely orbit. What does the original cube widget do when you drag on it? I believe the preceding axes widget had a similar to drag to rotate the axes.

zxq82lm:
> I like that this is part of UI now. However, dragging it around the screen is less useful from my perspective than being able to grab the cube or near the cube and drag to freely orbit. What does the original cube widget do when you drag on it? I believe the preceding axes widget had a similar to drag to rotate the axes.

---

- In the current Cube-Viewer (on main), holding right-click and dragging the mouse allows you to move the subject.
- The Cube-Viewer itself is supposed to remain fixed in the top-right corner. In this branch, holding left-click and dragging allows you to move the Cube-Viewer itself. If you prefer, we can also keep it fixed in the top-right corner?

PR #555, opened 2026-03-12
Perf/render server

zxq82lm:


x46085:
Looking like a lot of great exploration here!

Now that the big PR is in, I prefer we go for much smaller, targeted improvement PRs based on latest main, rather than a big PR. This will help us increment towards stability after so many changes.

PR #565, opened 2026-03-16
Remove old python extension files when building

anderspitman:
I was hit with a tricky-to-debug issue where the name of the extension file had changed (*.cpython-313-*.so to *.abi3.so), and the old one had import precedance. This removes existing .so during build so they are always up-to-date.


x46085:
is this something you just have to do once? I wouldn't put this into the just file if you delete it once and it's not a problem anymore; easier just to tell the team of the footgun and leave it out of code

PR #580, opened 2026-03-25
ci: tracy-based tests for elodin db.

zxq82lm:
## Summary
- Add `tracy` feature to elodin-db (port 8090) with tracing spans on 7 hot paths (packet handling, table sink, component write, mmap append, follow stream, coalescing flush)
- Add `elodin-db-bench` binary that reproduces the customer scenario (400 components at 250Hz with a connected reader)
- Add Buildkite `perf-elodin-db` step in a new `:racehorse: performance` group
## Context
Customer reports high CPU on Aleph with ~400 component streams at 250Hz (issue #574). elodin-db had zero instrumentation and zero perf tests. This PR adds the foundation to measure, profile, and prevent regressions.

---

Buildkite, *performance* group:
image
Buildkite,  *performance* > *perf-elodin-db*: 
image


zxq82lm:
the current PR content is totally in this PR now: https://github.com/elodin-sys/elodin/pull/581

PR #594, opened 2026-04-10
fix/debounce-wake-all

zxq82lm:
## Summary
- Debounce `update_max`/`update_min` calls in `DBSink`: instead of triggering `wake_all()` on every `apply_value` call, accumulate max/min timestamps locally and flush once after the batch completes
- First optimization identified by the elodin-db profiling instrumentation — eliminates ~800 redundant `wake_all()` per tick in a 400-component batch (~30% of tick time)


x46085 (approved):
Awesome. Looks great, excited for this.

Recommend a heavy local testing pass on this:
- rc-jet example live
- rc-jet example recorded
- rc-jet example followed

Looking for weirdness, especially around the jump to end / real time playback mode

PR #596, opened 2026-04-12
fix: reject component names containing whitespace

RCSharm07:
## Summary

- Validates component names in `PyComponent::new()` and the `name` setter
- Names containing whitespace now raise `ValueError` with a message suggesting `snake_case` or `dot.separated` alternatives
- Empty names are also rejected
- Added unit tests for the validation logic

## Context

Addresses #255. The issue notes that whitespace in component names caused problems in the editor (PR #236 had snake_casing workarounds). This fix catches the problem at registration time instead of papering over it downstream.

## Changes

**`libs/nox-py/src/component.rs`:**
- Added whitespace and empty-string checks in `PyComponent::new()` (the Python API constructor)
- Changed `name` field from `#[pyo3(set)]` to `#[pyo3(get)]` with a validated `#[setter]` method
- Added tests for valid and invalid component names

PR #599, opened 2026-04-14
fix: reject component names containing whitespace #596

### LINKED ISSUE #255: Fail fast on component names with whitespace
In relation to PR #236, there was snake_casing going on in editor. But we should confirm at component registration that the names do not allow for whitespace. Perhaps a discussion on what should be permitted in component names.

zxq82lm:
- Based on  [#596](https://github.com/elodin-sys/elodin/pull/596).
- Original work is by @RCSharm07.


shanecelis:
I opened [a PR](https://github.com/RCSharm07/elodin/pull/1) against the #596 PR so that we can accept the original PR with all the changes include herein. Closing this PR.

PR #604, opened 2026-04-15
Add Hamann-Chen-line crate and wire curve compression into plot / 3D line.

### LINKED ISSUE #593: 3d Line Max Point Count Limitation
# Problem

Should show complete line but currently only shows up to a fixed maximum count of points which means the line is only shown up to some time $T$. This shows up as a line thats stop at some predetermined point. This is probably exacerbated by the switch to a higher telemetry rate: 4 times the telemetry rate will make the new line cut off time happen at T/4, or in general $Y$ times the telemetry rate will mean the line cut off time is $T' = T/Y$.

# Solution

1. Only show the latest M points. Con: The history goes away. 

2. Compress the point data for the first $X$ percentage of points. Con: Loses fidelity but shows complete range. 

3. Compress the entirety of the points once they cross the max threshold.

For compression one can use [Hamann Chen (1994)](https://web.cs.ucdavis.edu/~hamann/HamannChen1994a.pdf) or any other suitable algorithm. I have [implemented this algorithm before but in C#](https://gist.github.com/shanecelis/2e0ffd790e31507fba04dd56f806667a) which could be used as a reference. Here are the results one can expect from the algorithm:

Image

zxq82lm:
## Summary

Adds **`hamann-chen-line`**, a small Rust crate implementing **Hamann–Chen (1994)**-style polyline / time-series simplification (returns **indices** into the original data). The Elodin **Editor** uses it to **downsample very long live telemetry** so graphs and 3D trails stay within memory and GPU limits.

Also hardens **GPU index strip** generation (bounds-safe writes, better sampling when there are many chunks, and coordinated step scaling for **3D lines** so trails are not clipped before the latest sample).

---

## How `line_3d` uses the crate

A `line_3d` in the schematic is three **`Line`** assets (X, Y, Z) fed from the same component with **aligned timestamps**.

When a plotted component has **exactly three** lines and timestamps **match across all three**:

1. Build a 3D polyline \((x_i, y_i, z_i)\).
2. Call **`select_polyline3_indices`** from `hamann-chen-line`.
3. **Rebuild** all three `LineTree`s with the **same** timestamp array and **same** index set so the trail stays a coherent 3D curve.

If joint mode does not apply, each line is simplified alone using the 2D path below.

---

## How this affects **graphs** (2D)

Each scalar channel is one **`Line`** (time vs value). Simplification uses **`select_time_value_indices`**: same idea as a polyline, in the \((t, y)\) plane.

Optionally, **`keep_recent_fraction`** leaves the **newest** portion of the series uncompressed so recent detail stays sharp while older data is thinned.

---

### Tunable (Bevy `Resource`)

**`CurveCompressSettings`** (default from `PlotPlugin`, overridable at runtime):

| Field | Role |
|--------|------|
| **`enabled`** | Master switch for compression. |
| **`compress_after_total_points`** | Run only when total stored points exceed this. |
| **`compress_to_points`** | Target point count after a pass (the `n` passed into Hamann–Chen). |
| **`keep_recent_fraction`** | Fraction of the **tail** (by time) kept at full resolution; `0.0` = compress the whole series (still capped by `compress_to_points`). |

Compression is triggered from **`maybe_compress_all_graph_lines`** after graph/table packets update `Line` data (shared path for 2D graphs and the lines backing `line_3d`).

### Fixed / internal (not user-facing)

- **`CHUNK_COUNT`**, **`CHUNK_LEN`** — used to derive default thresholds in `CurveCompressSettings::default()` (capacity-style defaults for when to compress / target size).
- **`INDEX_BUFFER_LEN`** — max `u32`s per GPU index buffer for line strips. For **3D** extraction, if the strip would not fit, the renderer **doubles the sampling step in lockstep on X, Y, Z** (up to a bounded number of iterations) so indices fit and the trail still reaches the current end of the data.
- **`index_sampling_step`** uses an internal **`PER_CHUNK_OVERHEAD`** constant when estimating stride so very large chunk counts do not fall back to an unsafe step of `1`.

---

## Crate layout

- **`libs/hamann-chen-line/`** — algorithm + optional CLI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5222,6 +5222,7 @@ dependencies = [
  "fastrand 2.3.0",
  "flume",
  "fuzzy-matcher",
+ "hamann-chen-line",
  "heapless 0.8.0",
  "hex",
  "hifitime",
@@ -6760,6 +6761,15 @@ dependencies = [
  "crunchy",
  "num-traits",
  "zerocopy",
+]
+
+[[package]]
+name = "hamann-chen-line"
+version = "0.16.3-alpha.0"
+dependencies = [
+ "anyhow",
+ "clap 4.6.0",
+ "glam",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ resolver = "2"
 #
 # default-members = ["apps/elodin"]
 members = [
+    "libs/hamann-chen-line",
     "libs/elodin-editor",
     "libs/nox",
     "libs/elodin-macros",
--- a/libs/elodin-editor/Cargo.toml
+++ b/libs/elodin-editor/Cargo.toml
@@ -14,6 +14,7 @@ tracy = ["bevy/trace_tracy"]
 
 [dependencies]
 # math
+hamann-chen-line = { path = "../hamann-chen-line" }
 nox.path = "../nox"
 nox.default-features = false
 egui_material_icons = "0.5"
--- a/libs/elodin-editor/src/ui/plot/data.rs
+++ b/libs/elodin-editor/src/ui/plot/data.rs
@@ -67,16 +67,34 @@ pub struct CurveCompressSettings {
     /// - `0.2` — keep roughly the last 20% at full resolution; Hamann–Chen runs on the leading
     ///   prefix only.
     pub keep_recent_fraction: f32,
+    /// Non-destructive **archive + view** architecture. When `true`,
+    /// [`PlotDataComponent::push_value`] appends each accepted live sample to
+    /// [`LineTree::raw_slice`], and [`LineTree::compress_time_value_hamann`] reads from that
+    /// archive instead of the decimated view. Switching this off at runtime reverts to the
+    /// historical destructive pipeline (no archive, HC reads and rewrites the view).
+    pub archive_enabled: bool,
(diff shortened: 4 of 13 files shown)


shanecelis:
I wanted to see how this implementation looks compared to the original figure from the paper shown in issue #593, and it looks good.

sine_hamann_grid

I added this and it can be re-created by doing the following:
```sh
cd libs/hamann-chen-line/examples && make
```

shanecelis:
I'm exercising the rocket example with this change:
```diff
-    max_ticks=5000,
+    max_ticks=500000,
```
And I'm seeing some flickering:

https://github.com/user-attachments/assets/1750ffeb-d60b-4b64-b34f-4f4bbafd5c12

shanecelis (changes requested):
See what can be done about the flickering and what looks like an errant point. It's possible the flicker has something to do with big_space. (In the future, I wish to make it a cargo feature so I can turn it off just to see if issues like this are related to big_space.)

shanecelis:
The flickering issue has been resolved. Thank you, Benoît!

https://github.com/user-attachments/assets/bcfa7176-34d2-4716-a4de-95f76480131e

There is a gap artifact shown above, but it does not flicker, and it is a concern we can address later.

shanecelis (approved):

PR #612, opened 2026-04-24
chore: Add git hash suffix to elodin python version.

shanecelis:
Bring Elodin Python library to parity with rest of Elodin tools in showing version.

```sh
$ python -c "import elodin; print(elodin.__version__)"
0.16.3-alpha.0+c9175f29b*
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -33,7 +33,7 @@
       };
       elodin = rec {
         elodin-py = final.callPackage ./nix/pkgs/elodin-py.nix {
-          inherit rustToolchain;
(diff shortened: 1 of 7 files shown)


shanecelis:
Changed the dirty suffix from '*' to ".dirty" because that's [valid semantic version](https://semver.org). This changes elodin cli and elodin db to use the same ".dirty" convention.
```sh
❯ python -c "import elodin; print(elodin.__version__)"
0.16.3-alpha.0+a59dc4b74.dirty
```

x46085 (approved):

PR #614, opened 2026-04-24
chore: Add `embedded_lfs_asset!` to replace `embedded_asset!` macro.

shanecelis:
Check that the embedded bytes are not an LFS pointer file at compile time.


x46085 (approved):
fixes the ghost cube?

reminds me I think we still will have the ghost cube problem with nix derived builds, or does this fix it there as well?

shanecelis:
Yes, it will stop the ghost cube from happening, which probably means it will break whatever build is causing the ghost cube. Then we can add the `git lfs pull` to the build that needs to be there.

PR #616, opened 2026-04-24
Fix release paths for Windows checkout

zxq82lm:
## Summary
- Rename baseline CSV files that used `>` in paths so Windows checkout can succeed.
- Normalize CSV comparison keys so Linux exports with edge names still match sanitized baseline filenames.
- Skip the PyPI publish step because v0.17.0 wheels were already published.


shanecelis (approved):

.github/workflows/release.yml
```diff
@@ -378,7 +378,7 @@ jobs:
           UV_PUBLISH_TOKEN: ${{ secrets.UV_PUBLISH_TOKEN }}
         run: |
           pip install uv
-          uv publish dist/*.whl --token "$UV_PUBLISH_TOKEN"
+          echo "PyPI package for v0.17.0 was already published; skipping."
```
x46085, line 381: this is not the way.

Move PyPI publish to the end of the release pipeline so it goes last, and re-release everything as v0.17.1

scripts/ci/compare_baseline_csv.py
```diff
@@ -76,7 +80,10 @@ def _collect_csv_files(root: Path, file_prefix: str = "") -> dict[str, Path]:
             file_name = path.name
             if file_prefix and not file_name.startswith(file_prefix):
                 continue
-            files[rel_path] = path
+            safe_rel_path = _windows_safe_rel_path(rel_path)
+            if safe_rel_path in files:
+                raise ValueError(f"CSV path collision after sanitizing for Windows: {rel_path}")
+            files[safe_rel_path] = path
```
x46085, line 86: i looks like you fixed all the filenames, why do you need this code now?

x46085:
wtf guys

.github/workflows/release.yml
zxq82lm: I already removed the v0.17.0 tag (in the goal to re-run release.yml with the v0.17.0 tag).

scripts/ci/compare_baseline_csv.py
zxq82lm: Because the checked-in baseline filenames are sanitized now, but the CSVs exported during CI can still be produced with the original edge names containing `>` on Linux/macOS.

zxq82lm:
> wtf guys

Sorry the branch were in auto-merge...

PR #618, opened 2026-04-24
Fix leaking entities on schematic clear

### LINKED ISSUE #605: Old Schematic Entities Persist After Unload
# Problem

Schematic entities persist after a load of a blank schematic. These viewports have been hollowed out, but the entities remain and the `WorldPos` continues to be called which will cause performance degradation. This was found when testing PR #603, which created many viewports.

Image

# Solution

All top-level entities created by a schematic should be marked with component perhaps like this:

```rust
#[derive(Component)]
pub struct SchematicSource(pub Handle<SchematicDocumentAsset>);
```
When a schematic is unloaded, all entries matching its handle ought to be despawned. I believe we're currently playing whackamole where we try to despawn them on a per type basis. Let's systematize this and make it behave regularly and uniformly regardless of what it represents in the UI.

# Verification

A headless, if possible, automated test ought to load a schematic, count the number of entities present before and after load, and compare that the number of entities is the same as before.

The cargo inspector feature can be used to manually verify the current behavior, but I would like it caught by a test to guard against regressions in the future.

anderspitman:
Fixes #605

A change was made to the solution proposed in #605. Instead of tying entities to a specific schematic, since the current code treats a schematic load as a full wipe anyway, we just nuke everything at that point. So technically the leak isn't fixed when the schematic is cleared, but when a new one is loaded.


libs/elodin-editor/src/ui/schematic/load.rs
```diff
@@ -1263,3 +1262,113 @@ pub fn show_document_load_failures(
...
+    fn load_schematic(app: &mut App, schematic: &Schematic) {
+        let mut system_state: SystemState<LoadSchematicParams> = SystemState::new(app.world_mut());
+        let mut params = system_state.get_mut(app.world_mut());
+        params.load_schematic(schematic, None, None);
+        system_state.apply(app.world_mut());
+        settle(app);
+    }
+
+    #[test]
+    fn loading_then_clearing_restores_entity_count_to_baseline() {
```
shanecelis: Nice test! Would be happy to have this use [test_case](https://docs.rs/test-case/latest/test_case/) to exercise more schematics to make it comprehensive.

libs/elodin-editor/src/ui/schematic/load.rs
```diff
@@ -108,6 +106,9 @@ fn apply_fallback_frame_to_panel(
 #[derive(Component)]
 pub struct SyncedViewport;
 
+#[derive(Component)]
+pub struct SchematicSpawned;
```
shanecelis, line 110: This is fine seeing as we have a one-schematic application at present.

shanecelis (changes requested):
Let's add a few more test cases to cover each of the KDL elements.

anderspitman:
@shanecelis let me know if this is what you had in mind. Verified that the tests catch individual regressions in the schematic elements. Not sure if it covers 100% of elements since I'm not as familiar with the KDL, but I think it's close.

libs/elodin-editor/src/ui/schematic/load.rs
```diff
@@ -1263,3 +1262,257 @@ pub fn show_document_load_failures(
...
+    #[test_case("schematic_tree" => (true, false, false) ; "schematic_tree")]
+    #[test_case("data_overview" => (true, false, false) ; "data_overview")]
+    #[test_case("inspector" => (false, false, true) ; "inspector")]
+    #[test_case("hierarchy" => (false, true, false) ; "hierarchy")]
+    #[test_case("tabs { data_overview }" => (true, false, false) ; "tabs")]
+    #[test_case("hsplit { data_overview share=0.5 data_overview share=0.5 }" => (true, false, false) ; "hsplit")]
+    #[test_case("vsplit { data_overview share=0.5 data_overview share=0.5 }" => (true, false, false) ; "vsplit")]
+    #[test_case("tabs { video_stream \"obs-camera\" }" => (true, false, false) ; "video_stream")]
+    #[test_case("tabs { sensor_view \"drone.scene_cam\" }" => (true, false, false) ; "sensor_view")]
+    #[test_case("tabs { log_stream \"fsw.log\" }" => (true, false, false) ; "log_stream")]
```
shanecelis, line 1419: Wow, I had no idea that test_case could be this expressive. Very cool!

shanecelis (approved):
This is great. Better than I requested.

PR #619, opened 2026-04-27
Fix manual docs deployment from GitHub Actions.

zxq82lm:
## Summary
This started because `v0.17.1` was not showing up on the live docs site at `https://docs.elodin.systems/releases/changelog/`, even though `docs/public/content/releases/changelog.md` on `main` already had the right content.
I tried to rerun **Deploy Documentation** manually from GitHub Actions, but the workflow completed without actually rebuilding and publishing the production docs. The production deploy steps only ran on `push` to `main`, not on `workflow_dispatch`.
This PR fixes that and tightens the deployment rules.
## What changed
- Manual `Deploy Documentation` runs from `main` now rebuild and publish the production docs.
- Manual runs from non-`main` branches do not deploy production.
- Production verification also only runs for `main` production deploys.
- PR preview docs now only run for changes under `docs/public/**`.
- Workflow-only PRs no longer create preview docs sites.
- Existing PR preview cleanup behavior is preserved for docs PRs when they are closed or merged.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -72,9 +72,9 @@ jobs:
           path: gh-pages
           token: ${{ secrets.GITHUB_TOKEN }}
 
-      # For main branch pushes or manual runs: Build and deploy production
+      # For main branch pushes or manual runs on main: Build and deploy production
       - name: Build production site
-        if: (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'workflow_dispatch'
+        if: (github.event_name == 'push' || github.event_name == 'workflow_dispatch') && github.ref == 'refs/heads/main'
         run: |
           cd docs/public
           echo "🔨 Building production site..."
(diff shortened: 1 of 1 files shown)


x46085:
What's the scenario you're trying to solve for here?

zxq82lm:
> What's the scenario you're trying to solve for here?

To make manual docs redeploys actually refresh production. The workflow already exposes Deploy Documentation as a manual action, but before this change it only deployed on push to main, so re-running it manually would pass without updating https://docs.elodin.systems/releases/changelog/ to the latest content like v0.17.1.

x46085:
> To make manual docs redeploys actually refresh production. 

Assuming you're meaning this doesn't work as described:
https://github.com/elodin-sys/elodin/blob/main/docs/public/DEPLOYMENT.md#manual-deployment

Okay cool, but you'll need to do a lot of testing for this. For example, why did this PR create a new PR docs site? You didn't make any docs changes, so that shouldn't have happened. You'll need to test:

- deploys PR docs when it's supposed to
- tears down PR docs when it's closed / merged
- doesn't deploy PR docs when it's not supposed to
- doesn't run a workflow and error trying to tear down PR docs when they don't exist

etc

x46085 (approved):
TYM!

zxq82lm:
- Manual deployment just after the merge: https://github.com/elodin-sys/elodin/actions/runs/25065263700
- changelog is ok now: https://docs.elodin.systems/releases/changelog/

PR #621, opened 2026-04-27
Feat: Add elodin-py to aleph image

shanecelis:
# Problem

Only elodin-db is installed on aleph. We want to be able to run the Elodin Python module to exercise tests on device.

# Solution

Include the Elodin Python module in the aleph image.

To check if it's present, do the following:

```sh
$ python
import elodin
```

### DIFF
--- a/aleph/flake.nix
+++ b/aleph/flake.nix
@@ -14,6 +14,13 @@
 
     jetpack.inputs.nixpkgs.follows = "nixpkgs";
     rust-overlay.inputs.nixpkgs.follows = "nixpkgs";
+    # Monorepo: parent is the Elodin flake root. For an `aleph/`-only tree use
+    # `github:elodin-sys/elodin` (root) instead of path:..
+    elodin = {
+      url = "path:..";
+      inputs.nixpkgs.follows = "nixpkgs";
+      inputs.rust-overlay.follows = "rust-overlay";
+    };
   };
   outputs = {
     self,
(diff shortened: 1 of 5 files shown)


aleph/flake.nix
```diff
@@ -125,7 +153,8 @@
     baseConfigurationModules =
       builtins.attrValues baseModules
       ++ builtins.attrValues fswModules
-      ++ builtins.attrValues devModules;
```
labrnth, line 128: to reduce the volume of the top-level flake, this elodin-py module could be added to the set of dev modules instead (`aleph/modules/aleph-dev.nix`)

aleph/flake.nix
```diff
@@ -158,7 +187,14 @@
       });
     }
     // rec {
-      nixosModules = baseModules // fswModules // stmModules // stmConfigurationModules // devModules // configurationPresets;
```
labrnth, line 161: to reduce the volume of the top-level flake, this elodin-py module could be added to the set of dev modules instead (`aleph/modules/aleph-dev.nix`)

aleph/flake.nix
```diff
@@ -14,13 +14,21 @@
 
     jetpack.inputs.nixpkgs.follows = "nixpkgs";
     rust-overlay.inputs.nixpkgs.follows = "nixpkgs";
+    # Monorepo: parent is the Elodin flake root. For an `aleph/`-only tree use
```
labrnth, line 17: it should be possible to re-use some of the existing infrastructure that is used for building elodin-db. for example, `pkgs/elodin-db.nix` gets `src` from `pkgs/common.nix`, which already brings in the repo root as a source path

x46085:
@labrnth great comments, Shane only got initial pass done on this, then left it for bigger fish. I’ll make sure to apply your ideas, that’s the right approach

PR #622, opened 2026-04-27
fix: Preserve frustum sources in inactive tabs.

### LINKED ISSUE #620: Frustum not shown if viewport is not visible
# Problem

If you have a viewport that creates a frustum and that viewport is behind a tab (not displayed currently), then that frustum will not be shown. 

## Viewport visible, frustum shown (GOOD)
Image

## Viewport not visible, frutum not shown (BAD)
Image

# Solution

A viewport that creates a frustum should have its frustum visible in any viewport.

## An Attempt
In trying to understand the way the code currently works, I tried altering frustum/mod.rs to always create the frustum. This works in that it creates the frustum regardless of visibility, but the frustum it creates differs depending on the viewport's visibility.

```diff
modified   libs/elodin-editor/src/plugins/frustum/mod.rs
@@ -259,14 +259,15 @@ fn draw_viewport_frustums(mut params: FrustumDrawParams<'_, '_>, mut commands: C
     for (camera_entity, camera, projection, global_transform, config, render_layer_lease) in
         params.main_viewports.iter()
     {
-        if !camera.is_active {
-            continue;
-        }
-
         let Some(config) = config else {
             continue;
         };
 
+        if !camera.is_active && !config.create_frustum {
+            continue;
+        }
+
+
         let Some(render_layer_lease) = render_layer_lease else {
             continue;
         };
```

zxq82lm:
## Summary
Fixes frustum rendering when the source viewport is hidden.

- Hidden source viewports can still publish their frustum to visible viewports.
- Inactive tab cameras use the active sibling tab rect for projection sizing, avoiding the previous `1x1` distortion.
- Cached source projections are reused as-is once available.

## Video
https://github.com/user-attachments/assets/b024d362-61b7-454b-903c-bfe968cb6bca


shanecelis:
The frustum is preserved; however, there appears to be a regression. It's hard to explain so I've done a screen recording so you can see it for yourself.

https://github.com/user-attachments/assets/bed1cefa-443d-46c8-be1c-388ccd8f09c7

shanecelis (changes requested):
See comment and video.

zxq82lm:
https://github.com/user-attachments/assets/3c921877-ca45-4c81-90e1-497e9a3d9927

zxq82lm:
- Intersection color is now defined at the source viewport.
- Intersection outline is smoother.
image

shanecelis (approved):
Looks great! Thanks for the follow ups.

PR #626, opened 2026-04-30
Add frustum overlays and intersection support for sensor cameras.

zxq82lm:
## Summary

- Add frustum settings to `world.sensor_camera()`:
  - `create_frustum`
  - `frustums_color`
  - `projection_color`
  - `frustums_thickness`
- Create editor-side sensor camera frustum sources.
- Render sensor camera frustums in viewports with `show_frustums`.
- Include sensor camera frustums in ellipsoid coverage and 2D projection.
- Add Inspector controls for sensor camera frustums.
- Show a frustum color dot on `sensor_view` tabs.
- Update the ellipsoid example.

## Ellipsoid example (video)

https://github.com/user-attachments/assets/b3a3e47f-58db-4686-8753-0ec7509922ec


x46085:
Well that's exciting!

Is it possible to toggle on and off the ellipsoid? The intended purpose of the sensor camera is to produce a image from a simulated actual camera device in the 3D world, so this device would never see any gizmos or debugging things, only every the 3D scene models and scenery, etc. Fine to have in there as long as we can keep them default off, lest we confuse people on the expected usage of the sensor camera.

zxq82lm:
This the new ellipsoid example:
- An onboard camera on the drone pointing toward its left wing.
- The ellipsoid is only displayed in viewports (there is a toggle to enable ellipsoid display for sensor cameras; it is disabled by default).
- There is one frustum originating from the viewport (left panel) and another from the sensor camera (right panel), both visible in the central viewport.
https://github.com/user-attachments/assets/b3a3e47f-58db-4686-8753-0ec7509922ec

x46085 (approved):
This is a pretty amazing upgrade for the sensor cameras, nice work!

PR #631, opened 2026-05-04
Enable bevy/debug feature when tracing

anderspitman:
The [bevy profiling docs](https://github.com/bevyengine/bevy/blob/main/docs/profiling.md)  say this well enable bevy debug names in tracy. I haven't noticed a different when casually comparing, but I think we should enable it in case there's something useful that I'm just not noticing now because I'm not looking for it.


x46085:
> The [bevy profiling docs](https://github.com/bevyengine/bevy/blob/main/docs/profiling.md) say this well enable bevy debug names in tracy. I haven't noticed a different when casually comparing, but I think we should enable it in case there's something useful that I'm just not noticing now because I'm not looking for it.

this doc is amazing, we should def link to it in our skill doc

x46085 (approved):

PR #633, opened 2026-05-05
StepContext read a component data value at a specific timestamp

x46085:
While all previous usages with the StepContext was reading and writing values in between simulation ticks, relative to the current tick, this was found limiting if you needed to read from data in the past. There's no performance implications for doing this, as the entire database is in memory during the request, so this was simple to add.


shanecelis (approved):
This offers a nice clean solution in the Python `post_step` function to look up data from the database. Nice work, Dan! I wonder how we could expose this functionality on the Jax side.

PR #652, opened 2026-05-19
Big space refactor refactor: No facade if possible.

shanecelis:
I was poking around your branch to make a feature out of "big_space", and it's an excellent start! It works! I do have a suggestion: Let's not make a big_space facade. I get the reasoning behind making a facade in spatial_fallback.rs so that we can suppress `#[cfg(feature = "big_space")]` from being peppered throughout the code. But to me preserving the facade is one of the things I'd like to get away from because I think that big_space API is confusing to look at. I kind of want this:

```rust
cfg_select! {
    feature = "big_space" => {
        // Big confusing chunk of code that deals with origins and cells and gunk.
    }
    _ => {
        transform.translate = f(x); // Oh, the above at heart just sets the position. 
    }
}
```

I don't claim that all my changes are strictly better, but I prefer to see that we simply omit systems instead of creating a stub that doesn't do anything. You can wield the `#[cfg(feature = "big_space")]` pretty freely as you'll see in my changes.

I know your branch is still in draft, so if you have local commits you want to push, feel free, and I'll re-base this branch if you want to accept its changes.


shanecelis:
I remove "big_space" from the default features while I was developing this. It's not intended to stay that way.

zxq82lm (approved):
LGTM into into chore/bevy-0.18-big-space.

PR #677, opened 2026-06-02
fix(editor): migrate env map ktx2 assets to git LFS.

### LINKED ISSUE #676: A couple broken git fls files
@zxq82lm  I think #647 had a minor breakage for LFS. When I try cloning a local repo to another location:

```
git clone elodin test
Cloning into 'test'...
done.
Encountered 2 files that should have been pointers, but weren't:
	libs/elodin-editor/src/assets/diffuse.ktx2
	libs/elodin-editor/src/assets/specular.ktx2
```

I git bisected it to commit 893ff69d1e4568222e07b40b23518aae71131604.

I think the problem is maybe that those files existed before, but the PR tried to move them to LFS? Not sure on that.

zxq82lm:
## Summary

PR #647 added `*.ktx2` LFS tracking in `.gitattributes` for skybox assets, but `libs/elodin-editor/src/assets/diffuse.ktx2` and `specular.ktx2` remained stored as git blobs (~1.5 MB / ~2 MB). Fresh clones reported:
```
Encountered 2 files that should have been pointers, but weren't:
  libs/elodin-editor/src/assets/diffuse.ktx2
  libs/elodin-editor/src/assets/specular.ktx2
```

These 2 files are used by `EnvironmentMapLight` on viewport cameras.

## What changed
Migrated `diffuse.ktx2` and `specular.ktx2` to LFS pointers via:
```
git lfs migrate import --no-rewrite \
  libs/elodin-editor/src/assets/diffuse.ktx2 \
  libs/elodin-editor/src/assets/specular.ktx2
```

## Documentation
- https://github.com/git-lfs/git-lfs/blob/main/docs/man/git-lfs-pre-push.adoc
- https://man.archlinux.org/man/extra/git-lfs/git-lfs-migrate.1.en

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 0cd9b8cef03948a8e379304d4a6d0864f5ccac70. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


x46085:
did you mean to wholly remove the lightmap? I think this is used when there's not a skybox configured no? 

I guess as an alternative you could create a really bland default skybox and use that as fallback

shanecelis:
I think what triggered the new warnings is just this one line in `.gitattributes`:
```
*.ktx2 filter=lfs diff=lfs merge=lfs -text
```
~~I'd try removing that and ensure you don't have any lfs pointers for these files. Ultimately it'd be preferable to have .ktx2s as LFS files, but the ones present are from a long time ago, so we'd have to rewrite the repo and do a force push to free that space, which I don't think is worth it right now.~~

EDIT: I think if we just making the old .ktx2 files LFS files, and it will be fixed. Using this command: `git lfs migrate import --no-rewrite --include="*.ktx2"`

x46085 (approved):

PR #684, opened 2026-06-05
s10 Driven in-process Monte Carlo

x46085:
- Solves the compounding memory problem of running multiple simulations 
- Provides tooling and scaffolding around designing and instrumenting a Monte Carlo sweep over simulations

---

# Monte Carlo ergonomics + Apollo editor visuals

## What this PR is about

This branch has three independent pieces, all built around the Apollo lander example:

1. **Monte Carlo ergonomics** — make it easy for anyone to take a simulation and get a Monte Carlo campaign running in a few minutes.
2. **Thruster exhaust particles (KDL)** — GPU exhaust plumes for the lander, declared in the scene file and driven by live sim data.
3. **View-cube picking fix** — the orientation cube stays clickable even over the huge landing-site terrain.

Each part is described below.

---

# Part 1 — Monte Carlo

## How Monte Carlo works

Monte Carlo = run the *same* simulation many times, each time with slightly different input values, then look at the spread of results. It tells you how robust a design or controller is, not just whether it works once.

In Elodin the flow is:

1. **Declare your inputs.** In the sim you mark the values you want to vary as `el.monte_carlo.Param(type, default=..., min=..., max=...)`.
2. **Write a sampling spec (`spec.toml`).** This says how to pick values for each input: a fixed number, or a random `uniform` range, etc. It also says how many runs (`n_samples`).
3. **Sample into a plan (`plan.csv`).** The spec is turned into a concrete table: one row per run, one column per input value. This is the source of truth for what each run actually used.
4. **Run the campaign (`campaign.toml` + `elodin monte-carlo run`).** A worker pool runs all the rows in parallel. Each run gets its own database and ports.
5. **Score each run with hooks.** A `post_run` hook decides if a run "passed" (for Apollo: did it land softly?). A `post_campaign` hook runs once at the end (for example, to summarize results or fail CI).
6. **Read the outputs.** `results.csv` (one line per run, with a `passed` column), `summary.json` (totals), plus a human-readable report.

Two things worth knowing:

- **A run that "completes" is not the same as a run that "passed."** Exiting cleanly only means the process didn't crash. Whether it met the real goal (e.g. a soft landing) is decided by the scoring hook.
- **By default the campaign process exits 0, even if some runs failed.** That's on purpose: a real Monte Carlo expects a failure rate, so the process shouldn't error out. If you want CI to fail on any failure, you add a `post_campaign` hook that raises when the failure count is above zero.

## The `quickstart` command

```bash
elodin monte-carlo quickstart path/to/main.py campaigns/my-sim
```

For any sim that declares its inputs, this reads them and writes a starter campaign you can edit:

```text
campaigns/my-sim/
  spec.toml        # one line per input: random range when min/max are set,
                   # otherwise a fixed value
  campaign.toml    # worker pool + which hooks to run
  hooks/score.py   # post_run: marks a run "passed" (edit this for your goal)
  hooks/gate.py    # post_campaign: fails the process if any run failed (for CI)
```

You then edit the ranges and the pass rule, and run the `elodin monte-carlo run` command it prints for you.

## Shared metric helpers (`hooks/mc_metrics.py`)

The Apollo hooks and scripts share their metric logic in one file, `examples/apollo-lander/hooks/mc_metrics.py` (`soft_landing`, `traj_rmse`, `run_passed`, plus small JSON/number helpers). It is imported by `score.py`, `report.py`, `ci_gate.py`, `ci_score.py`, and `calibrate.py`, so the pass/fail criteria are defined in exactly one place. The hook runner adds the hook's own folder to the import path, so a hook can import a sibling helper file no matter which directory you run from.

## Calibration loop (`calibrate.py`)

`calibrate.py` is the optional automated tuning loop: it looks at a campaign's results, finds the best run, narrows the input ranges around it, and re-runs. Notable details:

- It reads the best run's input values from `plan.csv` — the table of what each run actually used.
- It only considers runs that met the real goal (soft landing), not runs that merely finished without crashing.
- When it narrows a range it stays inside the original range, so the search only ever tightens.

## CI

Two fast Buildkite steps guard the Monte Carlo path on every push (both target a few minutes, not a full campaign):

- **`monte-carlo apollo`** runs a single nominal sample on a shortened Apollo simulation, checks the run produced its output, and fails the build only if a run failed (via the `post_campaign` gate hook). This exercises the real campaign + SITL plumbing end-to-end.
- **`monte-carlo quickstart`** scaffolds a campaign from the lightweight `examples/monte-carlo` sim, then runs the generated artifacts unedited (2 samples). It asserts every run passed the generated score hook, so it proves the scaffold is actually runnable: param extraction, spec/campaign validity, and the generated hooks resolving and passing.

---

# Part 2 — Thruster exhaust particles (KDL)

The Apollo lander now shows GPU exhaust plumes in the editor, declared in the scene file (`apollo-lander.kdl`) and driven by live simulation data. The plugin provides two reusable effect presets (`plume`, `cold_gas`); all engine placement, direction, and channel wiring live in KDL — nothing per-vehicle is hard-coded in Rust.

How it works:

- Each engine is a `thruster` node attached to the lander object in the KDL scene. The node says where the engine is, which way it points, and which sim channel controls how hard it fires.
- The **descent engine (DPS)** uses a vector channel (`lander.main_thrust_viz`): both the direction and the strength of the plume follow the thrust vector.
- The **16 RCS attitude jets** each use a scalar channel (`lander.rcs_thruster_viz[i]`) as a "cold gas" puff. Their order matches the thruster axis/sign tables in `sim.py`.

In code this is a new editor plugin (`thruster_particles`) plus the KDL parsing and serialization for `thruster` nodes, and a small rim-glow material/bloom for the engine and terrain look. The effect is editor-only and does not affect the physics.

---

# Part 3 — View-cube picking fix

Over the Apollo scene the cube could become unclickable, because the giant landing-site terrain mesh sat "in front of" the cube for the click test and absorbed the click.

The fix makes the cube's picking filter robust: it now keeps the cube active whenever **either** the cursor is physically over the cube's little overlay viewport **or** the UI input system reports the cube as the thing under the pointer. Previously it relied on a single signal that could briefly be empty (a one-frame timing gap), which let the big terrain mesh win the click. Using both signals together (an OR) closes that gap, so the cube stays clickable even over km-scale geometry.

---

# How to try the quickstart

`quickstart` builds a campaign from the params a sim **declares**. Some sims (like `examples/drone`) don't declare any yet, so you add them first.

**1. Declare the params in the sim.** Add a spec and feed it to the world build. For the drone, expose e.g. mass and a thrust gain in `examples/drone/main.py`:

```python
import elodin as el

SPEC = el.monte_carlo.params_spec(
    mass=el.monte_carlo.Param(float, default=2.586, min=2.0, max=3.2),
    motor_thrust_hover=el.monte_carlo.Param(float, default=0.689, min=0.5, max=0.85),
)

# `params` both answers `python main.py params` (read by quickstart) and
# provides the sampled values at run time.
params = el.monte_carlo.params(SPEC)
TALON_QUAD_CONFIG.mass = float(params.get("mass", 2.586))
TALON_QUAD_CONFIG.control.motor_thrust_hover = float(params.get("motor_thrust_hover", 0.689))
```

Score each run so the campaign has something to compare, e.g. in `post_step`:

```python
el.monte_carlo.result(final_altitude=..., upright=...)
```

**2. Scaffold the campaign** from the now-declared params:

```bash
elodin monte-carlo quickstart examples/drone/main.py campaigns/drone
```

> If you skip step 1, quickstart still runs but warns and writes an empty `[monte_carlo.variables]` — nothing to vary.

**3. Edit and run.** Tweak `campaigns/drone/spec.toml` ranges and the `hooks/score.py` pass criterion, then:

```bash
elodin monte-carlo run examples/drone/main.py \
  --campaign campaigns/drone/campaign.toml \
  --spec campaigns/drone/spec.toml \
  --out dbs/drone
```

See `examples/monte-carlo` for a complete sim that already declares params.

---

> [!NOTE]
> **Medium Risk**
> Touches process lifecycle (default reaping of elodin/elodin-db), port allocation, and JIT mmap constant lifetimes across workers; mitigated by serialized CI and explicit docs, but campaign collisions or memory bugs could affect local/CI runs.
> 
> **Overview**
> Adds a **native Monte Carlo stack**: workspace crate `libs/monte-carlo`, `elodin monte-carlo` CLI (`quickstart`, `template`, `sample`, `run`, `resume`, `report`), Python `el.monte_carlo` for params/result/context, and docs/CLI reference. Campaigns pin a shared `ELODIN_CACHE_DIR`, auto-size workers, reap stray `elodin`/`elodin-db` by default, optional `--memory-probe`, and run serialized **Buildkite** steps for Apollo + quickstart smoke tests.
> 
> Ships **`examples/apollo-lander`**: NASA telemetry + reference module, 6-DOF sim with kinematic truth ghost, Rust UDP “LGC” controller, campaign hooks, `calibrate.py`, KDL scene (GLB assets via LFS), and CI `*.ci.toml` + gate hooks.
> 
> **Cranelift** interns large StableHLO constants (>1 MB) into an mmap content cache (`DenseExternal` / import linkage) so parallel workers share resident memory; skills document cache lifetime and campaign usage.
> 
> **Editor/KDL**: `bevy_hanabi` **thruster** exhaust on `object_3d` (vector/scalar `intensity`, `plume`/`cold_gas`), viewport **bloom** + GLB **emissivity/glow**, and schematic docs for KDL booleans and thruster grammar.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 45c90fd31a234bb32ba171face8ae1424047419c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


zxq82lm:
This discussion: https://github.com/elodin-sys/elodin/discussions/687

zxq82lm:
On this branch / Apollo example, the 3D cube isn’t working well at all (big space?); I haven’t investigated it yet.

x46085:
> On this branch / Apollo example, the 3D cube isn’t working well at all (big space?); I haven’t investigated it yet.

I noticed that; very annoying. I did narrow it down to the imported landed site GLB, when you disable that the problem goes away. No idea why that mesh is problematic, but it’s clearly a failure case we should resolve robustly for future users.

danAtElodin (approved):

PR #692, opened 2026-06-18
Enhance the skybox creation command palette with options.

zxq82lm:
https://github.com/user-attachments/assets/d09b32e4-bd2d-4003-a18a-e6603772272b

---

> [!NOTE]
> **Low Risk**
> Editor-only command palette and documentation changes; generation still uses the existing `GenerateSkybox` path with optional fields that were already defined.
> 
> **Overview**
> **Generate Skybox** in the command palette is now a short wizard instead of firing generation right after the text prompt.
> 
> After you enter a description, you pick a **Blockade style** (M3 Photoreal, UHD Render, or Advanced), then a **local cubemap face size** (1K–16K, with **4K / 1024 px faces** as the default and first option so Enter without moving the selection matches the docs). The chosen `style` and `resolution` are sent on `GenerateSkybox` along with the prompt; existing guards (busy state, API key, plugin) still run on the final step.
> 
> Nested palette navigation **remembers filter text** when drilling into sub-pages and **restores it on Backspace** when you pop back (e.g. returning to the skybox description). Docs for the command palette are updated to describe the new flow and the 4K default.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f7d869c4b0b1ea3326391516fe78deabe0140c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/public/content/reference/command-palette.md
+++ b/docs/public/content/reference/command-palette.md
@@ -51,8 +51,8 @@ order = 8
 
 - `Skybox...`: submenu for skybox actions against `assets/skyboxes/manifest.ron` or
   `$ELODIN_ASSETS_DIR/skyboxes/manifest.ron`.
-  - `Generate Skybox...`: prompt for a description, generate a new skybox via Blockade, add it to
-    the manifest, and activate it.
+  - `Generate Skybox...`: prompt for a description, then choose a Blockade style and local cubemap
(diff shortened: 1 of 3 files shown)


x46085 (approved):
Ty!

PR #696, opened 2026-06-19
fix: vector-arrow rendering at planetary scale.

### LINKED ISSUE #693: vector-arrow cannot be seen at planetary scale
# Problem

`vector-arrow`s cannot be seen at planetary scale.

## Specific Instance

The `vector_arrow`s used in the new geo-frames example from PR #680 are supposed to draw the axes of rotation for each cube, but they are not visible. I have tried scaling them up and using greater thickness to no avail. The labels are present and in the correct positions.

Image

zxq82lm:
> [!NOTE]
> **Medium Risk**
> Editor gizmo/transform parenting changes affect all vector arrows; cranelift lowering changes can alter simulation numerics for large batched MLIR once pointer ABI is active—mitigated by new golden tests and docs.
> 
> **Overview**
> **Vector arrows** now render readably at very large scene scales (e.g. Apollo lander with `far=200000`). Arrow meshes run in **PostUpdate** after transform propagation, are **parented to the start endpoint** with rotation-only local transforms, and use **screen-pixel–based** head length, shaft length, and radius (with a minimum visible diameter when zoomed out). **`CachedArrowPose`** keeps the last good pose when endpoint reads fail briefly; visuals are **despawned** instead of hidden. The old `big_space` grid-cell delta path for arrow direction is dropped in favor of **`GlobalTransform`** on endpoints.
> 
> **Viewport zoom** is tied to the far clip plane via **`zoom_limits_for_far`**, updated when projection changes (inspector far edits and new cameras).
> 
> **cranelift-mlir (pointer ABI):** **`concatenate`** on `dim > 0` with multiple operands now **pairwise-folds** through `tensor_concat_nd_f64` (fixes wrong layout vs back-to-back `memcpy`); **`dot_general`** adds a **batched `matmul_f64`** fast path for leading batch dims and standard contracting layout. Docs and **`_mem`** golden tests cover both.
> 
> **apollo-lander.kdl:** DPS plume uses fixed body **direction** and scalar **`main_thrust_viz[2]`** intensity; removes an unused **window** layout block.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae61ae866ee2b660bea2ada8627634e777a6700c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

https://github.com/user-attachments/assets/ad9f9706-d9a1-45c5-af1c-675c966ab426

### DIFF
--- a/examples/apollo-lander/apollo-lander.kdl
+++ b/examples/apollo-lander/apollo-lander.kdl
@@ -70,7 +70,7 @@ object_3d surface.world_pos frame=ENU {
 // The Apollo LM GLB has four RCS quads; each quad has four perpendicular yellow nozzles.
 object_3d lander.world_pos frame=ENU {
     glb path=apollo-lunar-module.glb translate="(0, -2.5, 0)"
-    thruster name="DPS" effect="plume" body_frame=#true position="(0, -0.55, 0)" intensity=lander.main_thrust_viz emission_rate=180.0
+    thruster name="DPS" effect="plume" body_frame=#true position="(0, -0.55, 0)" direction="(0, -1, 0)" intensity="lander.main_thrust_viz[2]" emission_rate=180.0
     thruster name="rcs_0" effect="cold_gas" body_frame=#true position="(1.089, 0.870, -1.506)" direction="(0, 0, -1)" intensity="lander.rcs_thruster_viz[0]" emission_rate=1100.0 cutoff=0.006
     thruster name="rcs_1" effect="cold_gas" body_frame=#true position="(1.388, 0.874, -1.361)" direction="(1, 0, 0)" intensity="lander.rcs_thruster_viz[1]" emission_rate=1100.0 cutoff=0.006
     thruster name="rcs_2" effect="cold_gas" body_frame=#true position="(-1.180, 0.527, -1.369)" direction="(0, -1, 0)" intensity="lander.rcs_thruster_viz[2]" emission_rate=1100.0 cutoff=0.006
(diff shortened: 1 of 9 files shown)


x46085:
Very cool, glad you were able to figure out some of the cranelift library! You might inspect the ARCHITECTURE doc in the library folder and ensure it doesn’t need updating with the details of your changes, I was maintaining a list of supported MLIR operations in there. 

Also let me know when you’re 100% with this; I’ll need to test it against SDC to make sure it’s safe

zxq82lm:
> Very cool, glad you were able to figure out some of the cranelift library! You might inspect the ARCHITECTURE doc in the library folder and ensure it doesn’t need updating with the details of your changes, I was maintaining a list of supported MLIR operations in there.
> 
> Also let me know when you’re 100% with this; I’ll need to test it against SDC to make sure it’s safe

Thanks for pointing me to the ARCHITECTURE doc, I hadn’t thought to check it.

Sounds good, I’ll let you know once I’m 100% confident with the changes.

x46085:
Looking great!

@shanecelis is the axis arrow for ECEF correct? For some reason I thought it would be pointed directly at the center of the earth, but I admit I don't fully understand what ECEF looks like from orbit

zxq82lm:
@x46085 It's ready now. Let me know if you find something wrong.

shanecelis:
Yes, that looks right, but let's double check it. We can set our latitude to 90 degrees and longitude to 0 degrees. That should put us at the north pole, which should make the ecef Y-axis point in the same direction as NED, and voilà.

ecef

shanecelis:
Nice work fixing these arrows, Ben!

danAtElodin (approved):

PR #727, opened 2026-06-30
feat: db centric assets.

### LINKED ISSUE #719: DB-Centric Asset Architecture
This issue is investigating a re-architecture of how Elodin and Elodin-DB interacts with assets such as schematic kdl files, mesh glb files, skyboxes, and terrain data. Right now it's a confusing mess where things get loaded by sometimes the editor, sometimes the simulation, and sometimes copied or duplicated into the elodin-db as a fallback. This needs an overhaul in data flow.

Ultimate goal: 
- When a user runs a simulation, runs a SITL test, or deploys to an Aleph for HITL and telemetry, each creates a database, which should become a portable, replayable record of complete data and assets such that starting it and connecting an arbitrary Elodin editor to it from anywhere, will still reproduce the same experience.

Setting up a reproducible simulation or field test right now is very cumbersome. The inspiration for this major architectural change when I started investigating this plan @/home/dan/.cursor/plans/persist_kdl_schematics_in_db_1abfc4d8.plan.md  as a bandaid for KDL capture; this was in support of being able to deploy Elodin CLI and assets to an Aleph, which is partially supported now, but I discovered this gap in KDL coverage blocked me. After attempting, I realized that this is just another bandaid on a bigger architectural data problem, and we should address that instead.

Ideal case, I have an Elodin simulation project with the architecture:

.git
flake.nix
flight_software/
-- rust or cpp code, etc
simulation/
-- python_sim_code.py
-- main.py
-- assets/
---- skyboxes/
---- schematics/
---- terrain/

In this structure, I can run "elodin editor simulation/main.py" and launch a SITL simulation like the @examples/betaflight-sitl, which runs the Elodin simulation with a connected Elodin-DB, and the DB will be seeded with all the assets/ on launch, and "served" to the running simulation and connecting Editor. I can also set the ELODIN_ASSETS variable to point at the assets/ wherever it might be, which the database would respect on start.

Additionally in this structure, I can deploy from this Repository root to a target Aleph, including the Elodin CLI as described in the @aleph/template, and configure it to deploy this simulation and assets, and set the default database to also respect this assets/ folder.

A starting database should be able to find the assets folder to seed from the variety of places, first ELODIN_ASSETS, cwd, root alongside the sim file targeted, next to the main.py, etc

The key is the data flow is ALWAYS from source files to database to consumers in all cases. Gone is the code smell of loading assets into various places and trying to copy into the database and then fall-back serving them. 

We should be able to delete a lot of extraneous code and complexity. There are a lot of editor features to load and save KDL from file, interactive reload, etc that all go away. Saving or loading is always from the database via the asset server: @libs/db/src/assets_http.rs 

Draft proposal here:

[asset-architecture-plan.md](https://github.com/user-attachments/files/29367853/asset-architecture-plan.md)

@libs/db @libs/impeller2 @docs/public/content/reference @libs/elodin-editor

zxq82lm:
> https://github.com/elodin-sys/elodin/discussions/724 (rfd) & https://github.com/elodin-sys/elodin/issues/719 (issue)

---

# Current content

- [x] **Phase 0 — Ingest foundation**: copy the whole `assets/` tree into `{db}/assets/` once (`ingest_asset_dir`, `ELODIN_ASSETS`, `--assets`, copy-once guard). No consumer changes.
- [x] **Phase 1 — Schematics over HTTP**: editor loads the active schematic (`schematic.active`), the library, and window sub-schematics entirely over HTTP; `__index__` listing endpoint; drop the window filesystem fallback.
- [x] **Phase 2 — Editor write-back, named save & DB-native open**: `PUT` on the asset server (with a write gate); "Save" writes to the DB instead of disk; file dialogs and the `read_dir` picker are removed. "Save Schematic As..."
  stores under `schematics/<name>.kdl` (name validated to `[A-Za-z0-9_-]`) and "Save Schematic" overwrites the currently active key, so several named schematics coexist. "Open Schematic..." lists them via `GET /__index__/schematics/` and loads the chosen one over HTTP (`OpenDocumentFromActiveRequest`).
- [ ] **Phase 3 — Terrain over HTTP**: ingest `assets/terrains/`; convert the `bevy_world_mesh` loader to lazy per-tile HTTP fetch; delete terrain filesystem reads.
- [x] **Phase 4 — Consolidation & cleanup**: remove `ELODIN_ASSETS_DIR`, `ELODIN_KDL_DIR`, `schematic.content`/`schematic.path`; generalize follow-mode to a full-tree mirror via `__index__`. A single asset code path everywhere.
- [x] **Phase 5 — Aleph & deploy story**: wire `ELODIN_ASSETS` into Aleph's `elodin-db` service (ingest on each fresh boot); make `betaflight-sitl` the reference SITL example. A HITL DB becomes a complete, portable record.
- [x] **Phase 6 — Docs & tests**: rewrite `db-asset-server.md`, `schematic.md`, `elodin-cli.md` around the single namespace; tests (copy-once, traversal rejection, `__index__`, PUT round-trip, `Window` rewrite/collect, follow full-tree mirror).

---

> [!NOTE]
> **High Risk**
> Touches core DB lifecycle, follower asset sync, and schematic metadata migration; incorrect ingest or mirror behavior could affect portable recordings and replay, though copy-once and read-only follower gates mitigate data loss.
> 
> **Overview**
> Completes the **DB-centric asset model** (RFD #724): simulations and `elodin-db run` **ingest** a whole `assets/` tree into `{db}/assets/` once (`ingest_asset_dir`, copy-once guards so existing recorded DBs are never wiped), with **`ELODIN_ASSETS`** (replacing `ELODIN_ASSETS_DIR`) and CLI **`--assets`**. Schematics live as stored assets; metadata uses **`schematic.active`** instead of inline **`schematic.content`**, with KDL paths rewritten to **`db:`** on ingest/upload.
> 
> The **DB Asset Server** gains **`GET/PUT`**, **`/__index__`**, schematic upload rewrite, **`assets.revision`** bumps, and **`SubscribeLastUpdated`** config pushes. **Followers** mirror the full tree via the index (single-flight coordinator, retries, safe prune) and reject client asset/config writes on read-only mirrors.
> 
> **Aleph** wires ingest into the `elodin-db` NixOS module (`assetsDir`, seed ordering, tmpfiles). **Headless** and editor paths load/reload the active schematic over HTTP keyed on **`assets.revision`**. Docs and examples (`betaflight-sitl`, portable DB replay) are updated to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4cffc613b74b8a37472e5d5fc15034e46c67c10b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


x46085 (commented):
Don’t ever fall back. The repo has a rule “never create backwards compatibility at the cost of complexity (ie extra code). Ignore bugbot comments that say “oh no old databases will break”

x46085 (approved):
Tested on the Aleph, confirmed working with some follow up items I'll take on in a follow on PR.

PR #736, opened 2026-07-14
fix(editor): Stop line_3d flicker from f32 ECEF precision loss.

### LINKED ISSUE #735: The line_3d flickers even when paused.
# Problem

The line_3d element flickers a lot. Because this is generated from data we have, it hurts confidence in any data we represent. This may be due to the way we decimate the line.

## Screencast during Play

https://github.com/user-attachments/assets/ccb63ebc-1c03-4306-9492-7f1ce7f51ca0

## Screencast during Pause

https://github.com/user-attachments/assets/bd3c0bf1-2c9b-4e10-a79e-035946f6134b

## Suggested Diagnosis

- Turn off decimation and truncate to max to see if the issue goes away. It should if it's a decimation issue.
- Determine how often decimation happens. It should happen when new data is available, and even in that case.
- Are we decimating our already decimated data? It's possible that we're decimating data more than once. In fact I think I suggested such a strategy so that we would not have to process the whole dataset whenever there was new data.
- We added the Hamann-Chen algorithm, but Anders suggested there was another algorithm that maybe preceded it for a similar purpose. I'd ask AI if such a thing is present in our codebase (Anders doesn't recall). Is it possible those two are decimation algorithms are fighting each other?

zxq82lm:
## Causes

The flicker came from two independent problems:

1. **Precision — the line flickered even when paused.** A framed `line_3d` such as `NAVEKFSTATE.POS_ECEF` uses Earth-scale coordinates (~6.4 million metres). Storing each point as `f32` snapped it to the nearest ~0.5 m, so the whole trajectory sat on a coarse grid and shimmered at the slightest camera move.

2. **Camera — the scene trembled (a bit) only while playing.** The viewport camera follows the vehicle every frame straight from the raw telemetry (`pos = vehicle + offset`, `look_at = vehicle`). Real telemetry is slightly noisy, so that noise was copied 1:1 onto the camera and the entire view (grid plus the otherwise-stable trajectory) appeared to shake during playback.

---

> [!NOTE]
> **Medium Risk**
> Touches viewport camera pose, plot ingest/query paths, and GPU line rendering for ECEF-scale data; behavior is well-tested but spans playback, historical load, and 3D/2D graph split.
> 
> **Overview**
> Fixes **line_3d** shimmer and playback view shake by addressing ECEF **`f32`** precision and follow-camera noise.
> 
> **ECEF trails:** Framed **`line_3d`** components now keep **raw** telemetry in **`lines`** for 2D graphs and a **`rebased_lines`** mirror with per-axis **`value_offset`** (`raw − origin` in **`f64`** before cast). Spawn sets **`LineFrameOrigin`**, **`GeoPosition`** at the frame reference, and ingestion via **`GetTimeSeries`** / live stream; **overview** is skipped when rebasing is configured (overview is already **`f32`**). The line shader subtracts **`world_origin`**; GPU index building uses joint XYZ writes with a **0.1 m** minimum segment spacing to collapse pad scribble.
> 
> **Follow camera:** Viewports with **`look_at`** get **`ViewportFollowSmoothing`** — shared exponential smoothing on **`pos`** / **`look_at`**, snapping only on first frame or large **raw** jumps (seeks), not smoother lag during fast flight.
> 
> **Rendering:** Sub-pixel line segments fade via **`smoothstep`** in **`line.wgsl`** to avoid **`normalize(0)`** flicker.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 121dc51448a945ca80353dbde40ee49a0bc645e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/elodin-editor/src/ui/inspector/viewport.rs
+++ b/libs/elodin-editor/src/ui/inspector/viewport.rs
@@ -463,21 +463,30 @@ fn eql_input(ui: &mut egui::Ui, editable_expr: &mut EditableEQL, ctx: &eql::Cont
 
 /// Time constant (seconds) of the follow-camera low-pass. Larger = smoother but
 /// laggier. Telemetry drives the follow pose only at the data rate (~5-10 Hz),
-/// so this must span a few data periods to actually reject the jitter; genuine
-/// fast motion is protected by the snap-to-target path, not by a small tau.
+/// so this must span a few data periods to actually reject the jitter. `pos` and
+/// `look_at` share the filter, so lag stays along-track and the vehicle remains
+/// framed; seeks jump straight via [`follow_should_snap`].
 const FOLLOW_SMOOTHING_TAU: f64 = 0.3;
+
+/// Minimum raw-target jump (metres) treated as a seek/teleport. Continuous
+/// mid-flight motion (Shane ~250 m/s → ~4 m/frame at 60 Hz) stays well below
+/// this; a timeline scrub of a second or more does not.
+const FOLLOW_SEEK_SNAP_MIN_METERS: f64 = 50.0;
 
 /// Per-viewport low-pass state for a following camera. Holds the previous
 /// frame's smoothed position and look-at target (viewport/sim coordinates) so
 /// telemetry noise in a `pos`/`look_at` EQL that tracks a moving entity does not
 /// shake the whole view during playback. Snaps to the raw target on the first
-/// frame and on jumps larger than the framing distance (seek / fast motion), so
-/// genuine motion is never lagged.
+/// frame and when the *raw* telemetry jumps farther than a seek threshold —
+/// not when the smoother merely lags during fast flight (that used to thrash
+/// against the framing distance and re-inject noise).
 #[derive(Component, Default)]
 pub struct ViewportFollowSmoothing {
     initialized: bool,
     pos: DVec3,
     look_at: DVec3,
+    prev_raw_pos: Option<DVec3>,
+    prev_raw_look: Option<DVec3>,
 }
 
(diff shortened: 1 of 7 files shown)


shanecelis:
I'm still seeing a shimmer. I sent you a screencast.

shanecelis:
I know I've passed this back to you a few times after exercising it on Linux because that's where my DB was. I'm now exercising it on macOS, and it looks _great_! It's exactly what I want to see on macOS. @zxq82lm, can you do me a favor, and exercise this on Linux and see if the jitter is apparent? I have had some graphics card warnings that make my machine a little suspect. If you have any trouble running this against the DB and KDL I gave you, pull me in and let's chat about it or screen share.

EDIT: I may have spoken too soon. I ran using the standard KDL, which does not do any ECEF, and it did look fine, so there's a clue there but I don't understand how it's different yet. I wanted to test two things: 1. Does it jitter during play? 2. Does it jitter when paused and rotated? It didn't jitter at all for number 2, and number 1 was almost flawless. However, I didn't mean to run with the original KDL, I wanted to use the new one that uses ECEF. Suspecting that ECEF was a problem, I created a new NED line colored blue and pushed 1 meter east, which should not have any floating-point issues, and it exhibits the same issue. I will send you the KDL I used. Here is a video of it being played and being paused and rotated.

https://github.com/user-attachments/assets/87fee1d3-e1f1-4c2f-9e39-27c1ba836046

I also tried fixing the camera to a position (2,0,-2) looking at the origin, to see if the camera being moved was creating the issue, but the jitter remained even with a fixed camera.

PR #739, opened 2026-07-15
feat(editor): geo-position gauge & orientation gauge.

### LINKED ISSUE #738: Feature request: Spatial Gauge
# Problem

A monitor for a ECEF position is not very practical. The position is on the order of 6x10^6 for each x, y, z, and interpreting quaternions from its numerical elements remains an uncommon skill.

# Solution

Let's create a spatial gauge that the user can specify what coordinate system the data is in: ECEF, NED, and ENU. And the user can specify and change what coordinate system it is displayed in. This ought to be the first of many special purpose gauges that make the information easier to interpret. I am open to other visualizations, but what I had in mind is the position shown in X, Y, Z and perhaps a gimbal similar to an aircraft artificial horizon instrument to show the orientation of the quaternion.

Image

zxq82lm:
### Video

https://github.com/user-attachments/assets/caf4722b-2190-4da0-a8cd-32fcfbd2ed41

### geo-position gauge (*inspector*)
Geo-position gauge

### orientation gauge (*inspector*)
Orientation gauge

---

> [!NOTE]
> **Medium Risk**
> Touches schematic serialization, geo-frame transforms, and playhead EQL resolution; risk is moderated by broad unit tests and UI-scoped changes rather than simulation or auth paths.
> 
> **Overview**
> Adds two **EQL-driven schematic panels** for live telemetry: **`geo_position_gauge`** (three converted coordinates in NED/ENU/ECEF/LLA) and **`orientation_gauge`** (a 2D attitude gimbal with optional reference quaternion).
> 
> Both panels share **`EqlBinding`** and a **`compile_gauge_exprs`** update pass (same playhead resolution as graphs/monitors), inherit **`source`** from global `coordinate` when omitted in KDL, and round-trip through KDL/WKT with explicit-source preservation on save. The position gauge reuses **component-monitor value cards** via extracted **`render_value_cards`**; **`bevy_geo_frames::ecef_to_lla_deg`** supports LLA display.
> 
> Wiring covers **tiles**, command palette, empty-layout shortcuts, inspectors, schematic load/export, and a **`rotating-cube`** example schematic. The orientation widget is a sizeable custom painter (frame triads, sky/ground fill, co-rotating axes) with extensive unit tests for parsing, frames, and shading.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e6c4fdd650670836145e15cb7f074ba357ef2c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


shanecelis:
Wow, looking good!

shanecelis:
I added a rotating-cube example shown in the video below. It does not need to stay in the examples directory. But I do want to draw your attention to it.

https://github.com/user-attachments/assets/0a198c3c-4012-4142-abb8-0aa1a302a2d5

Now I do like the black and white. But in this example, we have a cube that's rotating on the x-axis, and you'll see that the gimbal does not rotate as well, instead it hugs the smallest angle deflection, which makes it rock.

The ellipses inside the graphic are static. I had hoped they would be dynamic and show the planes of each of the axes. I wonder if it might be easier to show a sphere with a different texture for each coordinate frame.

Finally, when viewing this on the rocket example, I want the gimbal to show up as white, but the rocket standing up isn't at its identity orientation, so I feel like we need to allow the user to specify a reference quaternion in the kdl that by default is identity to permit that.

shanecelis (changes requested):
See comment for requested changes.

shanecelis:
This is looking great! You've put the D-axis in for NED. I tested this against ECEF at the north pole (lat = 90, long = 0), which is a good way to spot check things as it will give you the closest semblance to NED or ENU, and it is correctly rotating on all X-, Y-, and Z-axes. ECEF shown below is correct for identity quaternion at north pole.

north-pole

One minor issue I see is in the geo-spatial window the frame around the numbers expands to fill any available width. This isn't how monitors number frames work, and I don't prefer because it does not give a clear indication that the next elements will flow if you expand the window far enough even though they will. In this case I would try to make the frame around the numbers match the monitor number frame.

number-window

shanecelis (approved):
Everything I see looks correct now. Thank you!

PR #744, opened 2026-07-21
Feature: Add error covariance to ellipsoid.

shanecelis:
# Problem

We can show ellipsoids using spatial dimensions or a Cholesky decomposition of a error covariance matrix, but we can't accept the error covariance matrix directly.

# Solution

Accept the error_covariance directly.

# Validate

Use the new example to check its behavior.

```sh
elodin editor examples/covariance-ellipsoids/main.py
```

https://github.com/user-attachments/assets/dbf96ff5-6c33-4ac5-8577-4b9b0d645d05

---

> [!NOTE]
> **Medium Risk**
> Changes ellipsoid math and frame handling in `object_3d` (user-visible 3D correctness) plus impeller2 Bevy entity parenting; covered by new unit tests but touches shared editor infrastructure.
> 
> **Overview**
> Ellipsoids can now be driven by a **symmetric 6-pack `error_covariance`** EQL field (alternative to `error_covariance_cholesky`). The editor factors **P** at runtime, applies the confidence scaling, and maps the shape through the object’s **`frame`** / schematic **`coordinate`** (replacing a fixed ENU→Bevy basis). **Cholesky wins** when both fields are present; driver choice follows **field presence**, so a bad compile does not fall through to scale or the other mode.
> 
> KDL/WKT serde, schematic docs, plot **SeriesStore** allowlisting, and the object_3d inspector gain the new property and UI to switch between scale, Cholesky, and symmetric covariance.
> 
> **`examples/covariance-ellipsoids`** animates equivalent Cholesky vs direct **P** side by side for manual validation.
> 
> Also in this PR: headless schematic load keeps **`Coordinate`** in sync; **DB component** entities parent under **`DbComponentsRoot`**; monitors under **`MonitorsRoot`**; viewport **infinite grids** spawn after the floating-origin root when `big_space` is on; minor rendering tweaks (plane **depth_bias**, terrain **`try_insert`**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a68565486535b98581ba6cdb3f3fc8fc2260e8d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/public/content/reference/schematic.md
+++ b/docs/public/content/reference/schematic.md
@@ -141,7 +141,8 @@ viewport hdr=#true {
         | c e f |
         The editor Cholesky-decomposes P at runtime. When both
         `error_covariance_cholesky` and `error_covariance` are set, Cholesky
-        takes precedence.
+        takes precedence based on field presence (even if the Cholesky
+        expression fails to compile).
       - `error_confidence_interval` (default `70`) the percentage that if this
         were repeated 100 times, we would expect that in 70 cases, the true
         value would be within the bounds. In practice this means that the
--- a/docs/public/content/reference/schematic.md
+++ b/docs/public/content/reference/schematic.md
@@ -133,6 +133,15 @@ viewport hdr=#true {
         | a 0 0 |
         | b c 0 |
         | d e f |
+      - `error_covariance` as an alternative to Cholesky, specify the
+        symmetric covariance matrix P directly as a 6-pack. Example
+        `"(a,b,c,d,e,f)"` which describes:
+        | a b c |
+        | b d e |
+        | c e f |
+        The editor Cholesky-decomposes P at runtime. When both
+        `error_covariance_cholesky` and `error_covariance` are set, Cholesky
+        takes precedence.
       - `error_confidence_interval` (default `70`) the percentage that if this
         were repeated 100 times, we would expect that in 70 cases, the true
         value would be within the bounds. In practice this means that the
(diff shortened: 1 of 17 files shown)


zxq82lm (approved):
LGTM.

I started by trying to understand the math of this branch.
Then I asked the AI to do a review. One point stood out as *possibly* an issue,
and an easy fix:

#### Fragile coupling on `Single<Root>` marker entities

**Issue.** Three critical, high-frequency systems now depend on a marker entity
being present in *exactly one* copy:

- `WorldSink` → `Single<DbComponentsRoot>` — all telemetry ingestion
- `LoadSchematicParams` → `Single<MonitorsRoot>` — the whole schematic-load pipeline
- `TileLayout` → `Single<MonitorsRoot>` — tile rendering, every frame

**Risk.** In Bevy 0.19, if a `Single` query matches zero or more than one entity,
the parameter fails validation and the **entire system is silently skipped** (or
panics when driven via `SystemState`). So if a root is ever missing (plugin
order, teardown on hard-clear/reconnect) or duplicated (plugin added twice,
respawn), a whole subsystem stops — with no obvious error. The blast radius is
large relative to the feature: a purely cosmetic hierarchy-grouping detail can
take down telemetry ingestion or the tile UI.

**Why it's only a "maybe".** It works today because each root is spawned once at
`Startup` and never despawned, so there is always exactly one. This is an
implicit invariant with no safety net.

**Fix.** Make the dependency tolerant instead of load-bearing:
```rust
// instead of: db_components_root: Single<..., With<DbComponentsRoot>>
// use:        db_components_root: Option<Single<..., With<DbComponentsRoot>>>
```
May be others ways to consolidate.

x46085 (approved):
LGTM - let's land this and get the release going

PR #747, opened 2026-07-22
Fix windows build

anderspitman:
> [!NOTE]
> **Low Risk**
> Localized startup/rendering defaults and lockfile bumps; shadow resolution reduction may slightly soften directional shadows on high-end GPUs but improves stability on memory-constrained devices.
> 
> **Overview**
> Fixes **Windows build / startup** issues tied to graphics memory and dependency versions.
> 
> **Directional light shadow map:** On native (non-wasm) builds, the default `DirectionalLightShadowMap` size drops from **8192 → 2048**. Bevy allocates the depth texture even before any shadow-casting light exists; at 8192 that was ~256 MiB per layer (~1 GiB with four cascades), which could blow the graphics budget on lower-memory GPUs and **invalidate the render device** (called out explicitly for Windows-class hardware).
> 
> **Lockfile:** `gpu-allocator` and `chrono` now pull **`windows` / `windows-core` 0.62.2** instead of 0.57.0, keeping the Windows dependency graph consistent with the rest of the stack.
> 
> **KDL plugin API:** `fetch_active_schematic_kdl` is only **re-exported** from `kdl_document` on **non-wasm Unix** (matching the existing Unix-only `headless` consumer); the GUI path still uses it internally via `operations`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d17b054e0a051777d2e0c6bba0f23de20d80987. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7012,7 +7012,7 @@ dependencies = [
  "log",
  "presser",
  "thiserror 2.0.18",
- "windows 0.57.0",
+ "windows 0.62.2",
 ]
 
 [[package]]
@@ -7682,7 +7682,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.57.0",
+ "windows-core 0.62.2",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


x46085 (approved):
I hope this doesn’t make the shadows look ugly :-) you can check on the Apollo lander example, scrub to the end to see the lander shadow on the moon

PR #758, opened 2026-07-30
fix: View-cube zoom In/Out.

### LINKED ISSUE #755: bug: The cube’s Zoom In feature isn’t working properly.
- Holding down Option on macOS to turn the − into a + does not always work.
- The same issue occurs with Shift / Ctrl / Alt / Windows key / Fn on Linux.

Possible related issue: zooming in/out with the mouse wheel.

zxq82lm:
> This branch only fixes the issue with triggering the + button (Zoom In) in the cube viewer. It has been tested on macOS and Linux.
> Mouse-wheel zoom possible bug is not included.

## 2 root causes

- 1 **The icon belonged to the wrong viewport.** Each overlay creates its own icon material, but `spawn_viewport_action_buttons` published the handle in a global `ZoomIconMaterials` resource that was overwritten on every call. Since `spawn_view_cube_overlay` runs once per 3D tile, only the last created viewport could swap its icon, and none could once that tile was closed. The observed behaviour therefore depended on the tile layout and on the order in which tiles were opened, not on the operating system.

- 2 **The modifier was read on edges.** The system used `just_pressed` / `just_released` on `KeyCode::AltLeft` / `AltRight`. Any lost edge latched the button in the wrong mode until restart, and Alt is precisely the key window managers intercept on Linux for window dragging and Alt-Tab. The system was also registered in `PreUpdate` without any ordering against Bevy's `keyboard_input_system`, which clears `just_pressed` at the start of each frame.

> [!NOTE]
> **Low Risk**
> Editor viewport UX and input state only; behavior is covered by new unit tests with no auth, persistence, or network impact.
> 
> **Overview**
> Fixes view-cube viewport zoom so **Alt/Option** toggles zoom-in (`+`) vs zoom-out (`−`) from **held** modifier state instead of press/release edges, avoiding a latched wrong mode when the OS drops key events (Alt-Tab, focus loss, Linux compositor).
> 
> Zoom icons and action mode are **per overlay** via a `ZoomButtonIcon` component (replacing a single shared material resource), and `sync_zoom_button_mode` runs between input and picking so hover/click see the current mode. **Zoom in and out now share one ratio** (`VIEWPORT_ZOOM_STEP`), so consecutive `+` and `−` clicks cancel exactly; tests cover that and multi-viewport Alt sync.
> 
> **LogicalKeyState** clears all pressed keys on `KeyboardFocusLost` so modifiers cannot stay stuck after leaving the window without release events.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6ee743bf5fb0634f4fab890777b150b738698c4e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/elodin-editor/src/plugins/logical_key/mod.rs
+++ b/libs/elodin-editor/src/plugins/logical_key/mod.rs
@@ -1,4 +1,4 @@
-use bevy::input::keyboard::{Key, KeyboardInput};
+use bevy::input::keyboard::{Key, KeyboardFocusLost, KeyboardInput};
 use bevy::prelude::*;
 use std::collections::HashSet;
 
@@ -36,6 +36,7 @@ impl LogicalKeyState {
 fn update_logical_key_state(
     mut key_state: ResMut<LogicalKeyState>,
     mut keyboard_events: MessageReader<KeyboardInput>,
+    mut focus_lost: MessageReader<KeyboardFocusLost>,
 ) {
     key_state.just_pressed.clear();
     key_state.just_released.clear();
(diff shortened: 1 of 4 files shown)


anderspitman (approved):
Works for me on Linux and most of the code changes are tests.

@zxq82lm just to clarify, the ALT key works for me on main, but the Windows/Super key does not. This branch makes them both work. Does that sound right?

PR #762, opened 2026-08-03
fix(impeller2): Skip and log malformed fields instead of dropping the table.

zxq82lm:
## Problem

Among all component types, only `bool` can fail to decode: every numeric primitive is validity-free, so any byte pattern of the right length is a valid value, while a Rust `bool` is valid only for `0x00`/`0x01`. That single failure was both fatal and opaque — it aborted the whole table on every packet, and the bare `InvalidComponentData` named neither the component nor its type.

## Solution

The handling is generalized rather than spot-fixed: `apply` now skips a field whose bytes aren't a valid value, logs it, and carries on with the rest of the table. Every sink user (DB, nox-py, fsw, CLI, editor) gets this without a single line changed — there is no opt-in path and no `*_lossy` variant.

* `Error::InvalidComponentValue { component_id, prim_type, elems, buf_len }` carries the context that makes the log actionable; `realize_fields` remaps the decode failure to it.
* **Structural failures stay fatal.** A truncated table, a schema mismatch or a misaligned field means the packet or the vtable itself is broken rather than one component being bad, so those still fail the whole table. Only value errors are skipped, so no caller can persist a partial row believing it succeeded.
* Skipped fields are warned **at most once per offending component**, since a malformed field recurs on every packet and would otherwise flood the logs at telemetry rates. The dedup set is `std`-gated and process-wide; its lock is only taken on the error path, so the happy path is untouched.
* `tracing` is added with `default-features = false` so the core can report without requiring std; verified with `--no-default-features`.

## Tests

Two tests provoke the real failures rather than asserting on mocks:

* a table whose `bool` field carries `0x02` still delivers its other fields to the sink;
* a truncated table still fails with `BufferUnderflow` and applies nothing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8034,6 +8034,7 @@ dependencies = [
  "stellarator-buf",
  "thingbuf",
  "thiserror 2.0.18",
+ "tracing",
  "zerocopy",
 ]
 
--- a/libs/impeller2/Cargo.toml
+++ b/libs/impeller2/Cargo.toml
@@ -39,6 +39,8 @@ stellarator-buf.path = "../stellarator/buf"
 
 # errors
 thiserror = { version = "2.0", default-features = false }
+# no_std-compatible: lets the core report skipped fields without a std logger
+tracing = { version = "0.1", default-features = false }
 miette.version = "7.2"
 miette.optional = true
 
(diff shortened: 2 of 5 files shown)


x46085:
Why did you create a new "sink_lossy" instead of just upgrading the current sink? This implementation only saves the plot graph, other panels or sink users will still die.

x46085 (approved):

PR #766, opened 2026-08-07
Fix: Line_3d f32 quantization error

shanecelis:
# Problem

We had quantization error in line_3d still. @zxq82lm noticed it and created an example to reproduce it. It looks like this:

Screenshot 2026-08-07 at 5 02 10 AM

# Solution

Turns out my scheme of subtracting the first from the following points in the line was happening after the data had already gone through a f32 conversion. But it's now fixed and looks correct in ENU or ECEF:

Screenshot from 2026-08-07 02-08-57

---

> [!NOTE]
> **Medium Risk**
> Changes the 3D line extraction/render data path and depends on TelemetryCache coverage; falls back to f32 LineTree when cache misses, which could still show quantization in edge cases.
> 
> **Overview**
> Fixes **staircased `line_3d` trails** at ECEF-scale coordinates (~6.4e6 m), where the display path had already quantized samples to **f32** before the floating-origin anchor subtraction, so subtracting the first point could not recover sub-metre motion.
> 
> The render path now keeps **`LineSources`** (component, element, and EQL **`ElementAffine`**) on each line, uses **`collect_strip_timestamps`** aligned with strip decimation, and **re-reads f64 samples from `TelemetryCache`** for anchor and vertices. Subtraction stays in **f64** via **`anchor_local`**; only the small residual is cast to f32 for GPU buffers. **`sync_line_3d_anchor`** uses the same **`line_first_point_frame`** helper so entity pose matches the trail.
> 
> EQL parsing gains **`to_graph_component_affines`** so constant shifts/scales (e.g. `ball.pos_enu[0] + 1.5`) apply to trails, not only bare elements. Adds **`examples/f32-quant-repro`** (ECEF vs ENU side-by-side) and unit tests for timestamp alignment and ECEF smoothness vs pre-fix f32 behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c83314e9bdc0e2570366f0e3c8cbdc9451d0e24e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/f32-quant-repro/main.py
+++ b/examples/f32-quant-repro/main.py
@@ -214,5 +214,5 @@ if __name__ == "__main__":
         system(),
         simulation_rate=1.0 / SIM_TIME_STEP,
         max_ticks=MAX_TICKS,
-        db_path=db_path,
+        # db_path=db_path,
     )
--- a/examples/f32-quant-repro/main.py
+++ b/examples/f32-quant-repro/main.py
@@ -1,0 +1,218 @@
+"""Repro for f32 quantization in the editor's plot/line_3d display path (#760).
+
+A ball flies a smooth, slow climbing arc near a real launch site in ECEF
+coordinates (~6.37e6 m magnitude — the worst case for f32). The recorded
+trajectory is f64 and perfectly smooth, yet the ECEF `line_3d` trail
+staircases on screen: the display path casts each sample to f32 at ingestion,
+and at that magnitude one f32 ULP is ~0.25-0.5 m.
+
+The same motion is recorded a second time in a local ENU component (numbers of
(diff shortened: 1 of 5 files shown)


zxq82lm (approved):
Successfully checked. LGTM.

Image

---

These two unit tests should be updated:

- `f64_samples_keep_ecef_trail_smooth`
- `f32_samples_staircase_the_ecef_trail`

Neither touches the code the fix changed: they only exercise `anchor_local`, so both would still pass if `axis_values` went back to reading f32 from the `LineTree`. Testing `axis_strip_values_f64` against a populated `TelemetryCache` would guard it.

x46085:
Huge

PR #791, opened 2026-08-19
Fix betaflight takeoff

anderspitman:
betaflight wasn't taking off for me. The core fix was changing the gyro noise setting.

Added a few other changes as well:

* It wasn't running realtime at 2kHz, so I switched to 1kHz for now
* Added a python script that takes care of programming the EEPROM
* Instead of using `sed` to change the betaflight submodule to enable lockstep mode, passed a `-D` to `make` instead

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the betaflight-sitl example (sim tuning, build script, and docs); no production services or security-sensitive paths.
> 
> **Overview**
> Fixes Betaflight SITL liftoff by **lowering simulated gyro noise** in `sensors.py` to `1e-7` / `1e-8` bias drift (from much higher values that destabilized attitude and motors during bootgrace).
> 
> Sets the default **physics/PID lockstep to 1kHz** in `config.py` so the Python/UDP bridge can run in real time; docs and comments now treat high gyro/accel rates as **targets capped by** `simulation_rate`.
> 
> Adds **`init_eeprom.py`** to program and verify `eeprom.bin` (ARM aux, gyro LPF, `pid_process_denom`) over the CLI instead of manual `screen` only.
> 
> **`build.sh`** enables `SIMULATOR_GYROPID_SYNC` via `OPTIONS=` on `make` and drops `sed` edits inside the Betaflight submodule; debug builds use the same helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aefc2f50881e6427f027dedd6cdd94505784c66e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/betaflight-sitl/init_eeprom.py
+++ b/examples/betaflight-sitl/init_eeprom.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+
+import socket
+import subprocess
+import time
+from pathlib import Path
+
+root = Path(__file__).resolve().parent
+binary = root / "betaflight/obj/main/betaflight_SITL.elf"
+
+
+def receive_until(cli, marker):
+    response = b""
+    while marker not in response:
+        chunk = cli.recv(4096)
+        if not chunk:
+            raise ConnectionError("Betaflight closed the CLI connection")
+        response += chunk
+    return response.decode(errors="replace")
+
+
+def start_sitl():
+    sitl = subprocess.Popen(
+        [binary],
+        cwd=root,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+    )
+    for _ in range(50):
+        try:
+            cli = socket.create_connection(("localhost", 5761), timeout=0.2)
+            break
+        except OSError:
+            time.sleep(0.1)
+    else:
+        raise TimeoutError("Betaflight CLI did not start")
+    cli.settimeout(10)
+    cli.sendall(b"#")
+    receive_until(cli, b"# ")
+    return sitl, cli
+
+
+def command(cli, text):
+    cli.sendall(f"{text}\n".encode())
+    return receive_until(cli, b"# ")
+
+
+def initialize():
+    sitl, cli = start_sitl()
+    with cli:
+        for text in (
+            "aux 0 0 0 1700 2100 0 0",
+            "set gyro_hardware_lpf = NORMAL",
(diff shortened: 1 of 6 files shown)


examples/betaflight-sitl/sensors.py
```diff
@@ -93,7 +93,7 @@ def sample(self, m: jax.Array, bias: jax.Array, tick: jax.Array) -> jax.Array:
 # Note: Betaflight's attitude estimator is sensitive to noise during the
 # bootgrace/calibration period. High noise causes attitude drift and
 # motor imbalance at liftoff.
-gyro_noise = Noise(0, 0, 0.01, 0.001)  # Gyro noise + bias drift
+gyro_noise = Noise(0, 0, 1e-7, 1e-8)  # Gyro noise + bias drift
```
anderspitman, line 96: This is the core fix

x46085: Bingo. Great find.

x46085: Wait does this mean we’re no longer providing representative noise? Or the previous noise was unrealistic? I guess we’d need some truth data

x46085 (approved):

examples/betaflight-sitl/init_eeprom.py
```diff
@@ -0,0 +1,85 @@
...
+    with cli:
+        for text in (
+            "aux 0 0 0 1700 2100 0 0",
+            "set gyro_hardware_lpf = NORMAL",
+            "set pid_process_denom = 1",
+        ):
+            command(cli, text)
+        cli.sendall(b"save\n")
+        receive_until(cli, b"Rebooting")
+    sitl.wait()
```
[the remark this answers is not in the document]
anderspitman: Not a big deal and not worth bloating this convenience script. If this becomes a problem in practice we'll fix it later.

examples/betaflight-sitl/sensors.py
anderspitman: My read is that the previous noise was unrealistic. It was empirically tuned for 4kHz, but I think it's pretty surprising that it worked given the magnitude of the noise. This returns it much closer to the original noise. I do plan on keeping an eye on this though and want to tune it again once we have the performance up where we want it.

PR #796, opened 2026-08-21
fix: View-cube ecef face labels.

zxq82lm:
## Video
geo-frames example:

https://github.com/user-attachments/assets/e0d63f37-5aa4-476c-832e-83e8d3687c99

---

> [!NOTE]
> **Medium Risk**
> Touches view-cube transform, render-layer inheritance, and per-viewport spawning—easy to leak labels across cameras or fight Bevy extract if scheduling is wrong. Logic is well-tested and isolated to the overlay widget.
> 
> **Overview**
> View-cube face labels now stay **screen-horizontal** at any camera pose, including oblique ECEF views where 90° rolls left text ~60° off.
> 
> Each overlay camera gets its **own copy** of the labels on its own render layer (`KeepsRenderLayers`), so two viewports sharing a cube can spin independently without leaking copies into every view. `orient_face_labels_to_view` solves an exact in-plane angle after transform propagation and writes `GlobalTransform` so extract sees the spin.
> 
> The geo-frames example splits ECEF into equator and oblique viewports so this is easy to check. Overlay camera sync also skips no-op writes to cut change detection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51f3323dc0debcd798b9f7b556d03e7cc9ac4f23. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/geo-frames/main.py
+++ b/examples/geo-frames/main.py
@@ -112,9 +112,12 @@ def world() -> el.World:
         f"""
         coordinate frame=NED lat={LAT_DEG} lon={LON_DEG} alt={ALT_M}
         hsplit {{
+            vsplit {{
+                viewport name="ECEF Equator" frame="ECEF" pos="(0,0,0,1, 8000000,-80000000, 0)" look_at="(0,0,0,1, 0,0,0)" up="(0,0,1)" far=15000000.0 hdr=#true show_grid=#true active=#true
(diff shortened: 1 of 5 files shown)


x46085 (approved):
Wunderbar!

PR #814, opened 2026-08-24
test(apollo-lander): cover the descent reference profile

vyom-aggarwal:
apollo-lander shipped without tests, unlike the comparable falcon9 and voyager examples. reference.py drives both the in-sim truth ghost and the external guidance controller, and it carries two constants that its own comments document as having to match code elsewhere -- FOOTPAD_CONTACT_ALT_M with sim.FOOTPAD_HEIGHT_M, and TERMINAL_CONTACT_RATE_MPS with the controller's MIN_DESCENT_RATE_MPS -- with nothing enforcing either.

Add test_reference.py covering:

- sanity_check(), which cross-checks the derived SI measurements against the raw vendored CSVs but was previously only reachable by running reference.py as a script
- the profile endpoints: it picks the mission up at landing-radar lock-on (~11.8 km, ~800 m/s, ~107 km uprange) and ends stopped over the targeted site at footpad contact height
- both documented constant couplings, so drift on either side is caught

Altitude is deliberately not asserted to be monotonic: it is digitized from the mission-report chart and smoothed, so a handful of samples tick up by a few metres. The descent-rate sign is asserted instead.

---

> [!NOTE]
> **Low Risk**
> Tests-only addition for an example; no production logic, auth, or data-handling changes.
> 
> **Overview**
> Adds `test_reference.py` for the Apollo lander example, which previously had no tests around the shared descent truth profile.
> 
> The suite runs `sanity_check()` against vendored CSVs, asserts the profile starts at radar lock-on and ends at footpad contact with zero horizontal speed, and locks `FOOTPAD_CONTACT_ALT_M` / `TERMINAL_CONTACT_RATE_MPS` to `sim` and the Rust controller so those documented couplings cannot drift. Descent-rate sign is checked instead of altitude monotonicity, matching the digitized chart.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c5113811c668fb5d40e15973fdc6ebff434442d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/apollo-lander/test_reference.py
+++ b/examples/apollo-lander/test_reference.py
@@ -1,0 +1,63 @@
+"""Regression tests for the shared Apollo 11 descent reference profile.
+
+`reference.py` reconstructs the truth profile that drives both the in-sim truth
+ghost and the external guidance controller. It carries two constants that its
+own comments document as having to match code elsewhere -- the footpad contact
+altitude (`sim.FOOTPAD_HEIGHT_M`) and the terminal descent rate
+(`MIN_DESCENT_RATE_MPS` in `controller/`) -- but nothing enforced either
+coupling, so a change on one side could silently desynchronise the truth
+display from the physics or from the guidance law.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+import reference
+
+
+def test_vendored_data_matches_derived_measurements():
+    """reference.py's own cross-check of the derived SI values against the raw CSVs."""
+    report = reference.sanity_check()
+    assert report["ok"], report
+
+
+def test_profile_flies_the_descent_to_footpad_contact():
+    ref = reference.build_reference()
+
+    # Picks the mission up at landing-radar lock-on: ~38,700 ft, still carrying
+    # the residual orbital velocity, ~107 km uprange of the target.
+    assert 11_500.0 < ref.altitude_m[0] < 12_500.0
+    assert 600.0 < ref.horizontal_speed_mps[0] < 1_100.0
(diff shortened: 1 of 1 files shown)


x46085:
I don't understand, why does this make the current example better? Your description is very wordy, saying not very much with a ton. Need to understand right up front what you're trying to accomplish here.

As for the apollo-lander example, I'd be way more interested in research of the simulation model for the mission. I'm very aware that it only approximates the actual mission, and much of the required physics modelling is missing in order to more accurate match the actual historical record. For example, the RCS thrusters can throttle full range and fire continuously in our sim, which doesn't match the reality of the actual capabilities of the lander. This is why our lander can take a more aggressive route to the surface and land safely with a lot more remaining fuel. The actual lander had almost no fuel left at landing.

In fact, before you work on this anymore, you are required to go watch [First Man](https://www.youtube.com/watch?v=FubpK1Tho6M)

Afterward I'd love to see an issue ticket after a ton of research that explains a full list of all the ways our current modelling could be improved to match the real event better.

vyom-aggarwal:
@x46085 Sorry for that, you're right. The honest answer is it doesn't make the example better. It guards a maintenance detail: reference.py, sim.py and the controller share two constants the comments say must match, with nothing enforcing it. Worth having eventually, but not what this example is for.

Happy to close it and put the effort into fidelity instead. The RCS point is a good starting thread.

Watching First Man, then I'll open an issue with the full modelling-gap list.

PR #824, opened 2026-08-27
Real bdx fixes

anderspitman:
2 fixes:

1. the shader fix is because naga compilation was failing for me after @zxq82lm's recent fixes

2. the OOM fix is because my nvidia GPUs have a small BAR heap which was triggering the proactive OOM check, which ironically caused OOM every time.

---

> [!NOTE]
> **Low Risk**
> Small, targeted fixes: unreachable-style shader fallback and relaxed wgpu budget pre-check; slightly less early OOM protection on other GPUs.
> 
> **Overview**
> Fixes two runtime/build issues: terrain tiling prepass shader compilation and spurious GPU OOM on some NVIDIA cards.
> 
> In **`refine_tiles.wgsl`**, **`try_reserve_children`** now ends with an explicit **`return -1`** after its atomic reservation loop. Naga does not prove the loop always exits through the inner returns, so the **`i32`** return type was failing validation; the fallback matches the existing failure path when the tile buffer is full.
> 
> In **`editor_wgpu_settings`**, **`instance_memory_budget_thresholds.for_resource_creation`** is set to **`None`** instead of **`Some(99)`**. Vulkan preflight can treat NVIDIA’s small BAR heap as device-local and reject large allocations (e.g. terrain atlases) while VRAM is still free; disabling the proactive threshold lets the driver report real OOM instead.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d8902f4fee846033f5dc02df057fdcf6bf243d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/bevy_world_mesh/src/terrain/shaders/tiling_prepass/refine_tiles.wgsl
+++ b/libs/bevy_world_mesh/src/terrain/shaders/tiling_prepass/refine_tiles.wgsl
@@ -42,6 +42,10 @@ fn try_reserve_children() -> i32 {
             return current;
         }
     }
+
+    // Naga does not infer that the loop can only exit via one of the returns
+    // above, so an explicit fallback is required to satisfy the i32 return.
+    return -1;
 }
 
 fn subdivide(tile: TileCoordinate) -> bool {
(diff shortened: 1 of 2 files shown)


shanecelis (approved):
I so appreciate the comments to code ratio in this PR.

PR #825, opened 2026-08-28
Fix: aarch64 cortex flag

shanecelis:
# Problem

The new version of rustc that we use emits a new flag `--fix-cortex-a53-843419` that the zig cross compile does not know or accept.

# Solution

When rustc emits this flag, which will only be on aarch64 builds, then we use the `lld` linker which respects this argument. We also use zig's C++ library. The x86 build should be unchanged.

---

> [!NOTE]
> **Medium Risk**
> Changes release musl linking for aarch64 only, but the new wrapper/LLD path is intricate and affects how C++ and CRT objects are supplied at link time.
> 
> **Overview**
> **Fixes aarch64 `*-linux-musl` release builds** when rustc 1.98+ passes `-Wl,--fix-cortex-a53-843419`, which Zig 0.16’s `zig cc` rejects.
> 
> Musl cross-compile logic moves into **`scripts/zig-musl-wrapper.sh`**. CI and `scripts/test-musl-build.sh` now install thin **`cc`/`c++` trampolines** that set `ELODIN_ZIG_*` and exec that script instead of inlining wrapper bash in YAML.
> 
> On **link** when the cortex flag is present, the wrapper **unwraps `-Wl,` args** and runs **`zig ld.lld`** (which accepts the errata flag). It **rewrites `-lstdc++`** to Zig’s cached **`libc++abi.a` / `libc++.a`** for deps like openh264. Compile (`-c`) still goes through **`zig cc`/`c++`** with the same filtering of Rust triples and self-contained CRT objects.
> 
> **`.cargo/config.toml` wiring is target-specific**: **`aarch64-unknown-linux-musl`** gets a linker path only (no `link-self-contained=no`, so rustc’s musl CRT stays in play for the LLD path). **`x86_64-unknown-linux-musl`** keeps **`link-self-contained=no`** and the prior zig-cc-as-linker behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d1249e92c97fb391aa499d36ac420a85d4df5f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -199,59 +199,39 @@ jobs:
 
           # Wrappers needed because cc-rs treats CC_*/CXX_* as paths, not shell
           # commands, so we cannot inline `-target <triple>` in the env var.
-          # The wrappers strip the `--target=<rust-triple>` that cc-rs appends
-          # to every invocation: Zig rejects it with "UnknownOperatingSystem"
-          # because it expects its own triple format (e.g. `aarch64-linux-musl`
-          # without the `unknown` vendor component). We then pass the correct
-          # Zig-flavoured `-target` ourselves.
-          # printf (no heredoc) ensures the shebang sits at byte 0 of the file
-          # — kernels require `#!` as the first two bytes for it to be honored.
+          # printf (no heredoc) keeps `#!` at byte 0 of each trampoline.
           wrap_dir="/opt/zig-cc-${env_target}"
           sudo mkdir -p "$wrap_dir"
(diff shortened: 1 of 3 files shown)


zxq82lm (approved):
LGTM.

Ran scripts/test-musl-build.sh on Linux with Zig 0.16.0 — both targets pass (x86_64-unknown-linux-musl, aarch64-unknown-linux-musl), so the ld.lld path does work end to end.
The rest of the review looks good.

PR #829, opened 2026-08-31
Fix: Plot line flicker.

### LINKED ISSUE #822: Bug: Graph Line Flicker
# Problem

Graph lines can sometimes flicker. This video shows the behavior. Fortunately, no points are displaced in the graph, so it's a low priority.

https://github.com/user-attachments/assets/961db704-39a1-4c3c-ac34-2ebbcd10c904

# Conjecture

I wonder if a shared GPU buffer is being passed around somehow between the three lines.

zxq82lm:
> [!NOTE]
> **Medium Risk**
> Touches GPU buffer lifecycle and render replay with correctness guards; wrong replay would show stale or cross-line geometry, though buffer-ID checks mitigate that.
> 
> **Overview**
> Addresses visible plot flicker when SQL/query plots auto-refresh and when the GPU buffer pool defers new allocations.
> 
> **Stable XY assets across refresh:** Query plots now pass prior series handles into SQL batch processing so each series **refills the same `XYLine` asset** via new `replace_points` (swap data, keep shard slots). `queue_load` reclaims or resets GPU allocations through `content_replaced` and `xy_takes_from_pool`, so refreshes do not allocate a new pool buffer every interval.
> 
> **Deferred upload fallback:** When value-buffer uploads are deferred or over the per-frame budget, the extract path can **replay last frame’s cached draw** instead of omitting the line, but only if `plot_draw_replay_allowed` confirms the line still owns the same value buffer IDs (avoids drawing another line’s data).
> 
> Tests cover replace/refill behavior, replay safety, and allocation gating; an ignored test documents a latent `LineHandle` hook issue (#822).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9729009ac5f5e4ae938da457619003885da89800. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/elodin-editor/src/ui/plot/data.rs
+++ b/libs/elodin-editor/src/ui/plot/data.rs
@@ -778,7 +778,6 @@ pub fn sync_plot_lines_from_series_store(
     if !(range_changed || enabled_changed || (gen_changed && due)) {
         return;
     }
-
     sync_state.last_range = Some(range_key);
     sync_state.last_generation = series_store.generation();
     sync_state.last_enabled = fetch_ids.clone();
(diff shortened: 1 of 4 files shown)


shanecelis (approved):
Looks good to me. I checked it against both a SITL and real db on Linux.

PR #831, opened 2026-09-01
chore: Add release dry-run after merge to main.

shanecelis:
# Problem

When we do a release, we run code that isn't part of our usual CI process. Often times this provokes errors that are only seen during release.

# Solution

On every merge to main, but not every PR commit, do a release dry-run that runs more of the release process but excludes announcing and package publishing to ensure that the release problems are caught earlier.

---

> [!NOTE]
> **Low Risk**
> CI-only workflow changes; no production code paths, though dry runs add heavier jobs on every main push.
> 
> **Overview**
> Extends the **Release** workflow so merges to `main` (and manual `workflow_dispatch`) exercise the same artifact builds as a real release, without creating a GitHub Release or publishing to PyPI.
> 
> **Triggers and gating:** Adds `push` to `main` and `workflow_dispatch`, plus a `concurrency` group (in-progress runs on non-tag refs can be cancelled). **Publishing** is now tied to version **tags** (`startsWith(github.ref, 'refs/tags/')`) instead of “not a pull request,” so main pushes are not treated as releases. A new **`dry-run`** plan output is true for main/`workflow_dispatch` (not PRs, not tags).
> 
> **Build coverage:** `build-local-artifacts`, `linux-wheel`, `cargo-deb`, and `macos-wheel` run when `dry-run` is true as well as on tag publishes. **`host`**, **`announce`**, and **`pypi-publish`** still depend on `publishing`, so they stay skipped on dry runs.
> 
> **Status check:** A **`release-dry-run`** job runs after those build jobs and fails the workflow if any required build did not succeed, giving a single check for post-merge release validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3e3afd7f45a4c40caa332f2f2c65659b745e463. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,8 @@
 # * builds artifacts with dist (archives, installers, hashes)
 # * uploads those artifacts to temporary workflow zip
 # * on success, uploads the artifacts to a GitHub Release
+# * on push to main (or workflow_dispatch), dry-runs the full build
+#   without creating a GitHub Release or publishing to PyPI
 #
 # Note that the GitHub Release will be created with a generated
 # title/body based on your changelogs.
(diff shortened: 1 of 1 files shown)


zxq82lm (approved):
LGTM.
Good idea on merge.

PR #844, opened 2026-09-08
kdl: skip the bare timeline node when saving a schematic

### LINKED ISSUE #669: kdl: Skip bare timeline node when saving schematic with default settings.
## Problem
Saving a schematic always emits a top-level `timeline` node, even when all timeline settings are at their defaults:
```kdl
theme mode=dark scheme=default
timeline
skybox name=seaport
...
```

A bare timeline node (no `played_color`, `future_color`, or `follow_latest`) is functionally identical to omitting the node entirely — defaults apply on load. This adds noise to saved KDL files and can confuse users reviewing schematics.

## Suggested fix
Either:
- In `tiles_to_schematic`: set `schematic.timeline = None` when settings match `TimelineConfig::default()`, or
- In `serialize_schematic`: skip pushing the timeline node when `serialize_timeline` would produce an empty node.

Mohit-Ak:
Saving a schematic from the editor always wrote a top-level `timeline` node, even when every timeline setting was at its default. `serialize_timeline` already omits each individual property when it matches the default, so in that case the node came out completely bare:

```kdl
theme mode=dark scheme=default
timeline
skybox name=seaport
```

A bare `timeline` node is indistinguishable from no node at all — the loader does `schematic.timeline.clone().unwrap_or_default()` and `parse_timeline` falls back to the same defaults for every missing property — so it was pure noise in saved files.

This takes the second of the two options suggested in #669: `serialize_schematic` now builds the timeline node as before and only pushes it if it has at least one entry. That keeps the decision next to the code that already knows which properties are default-valued, rather than duplicating a `== TimelineConfig::default()` comparison in the editor's `tiles_to_schematic`. Non-default settings (`played_color`, `future_color`, `follow_latest`, `range`) serialize exactly as before.

Added `test_serialize_skips_bare_timeline_node`, which checks that a default `TimelineConfig` produces no `timeline` node, that the result parses back to default settings, and that a single non-default field (`follow_latest`) still emits the node.

Verified locally on the pinned toolchain (rustc 1.98.0):

```
cargo test -p impeller2-kdl        # 181 passed, 1 ignored (unit) + example_schematics ok
cargo fmt -p impeller2-kdl --check  # clean
cargo clippy -p impeller2-kdl --all-targets -- -Dwarnings   # clean
```

The new test fails on `main` with `default timeline settings should not emit a node: timeline` and passes with the change.

Fixes #669

---

> [!NOTE]
> **Low Risk**
> Serialization-only change with preserved parse behavior; saved files may lose a redundant timeline line without changing editor playback settings.
> 
> **Overview**
> Saving a schematic no longer writes a top-level **`timeline`** node when every timeline setting is at its default. **`serialize_schematic`** still builds the node via **`serialize_timeline`** (which already drops default properties), but only emits it if at least one property remains—so explicit **`range=full`** and other default-equivalent inputs normalize to no node on save.
> 
> **Backward compatibility:** parsing an empty **`timeline`** (or omitting it) still loads the same defaults; re-saving those files drops the noise. Non-default **`played_color`**, **`future_color`**, **`follow_latest=#true`**, and trailing **`range`** values are unchanged.
> 
> Docs and the **`examples/db-client`** schematic drop illustrative bare **`timeline`** lines; the schematic reference also documents save/load behavior and corrects the default **`played_color`** name to **`yalk`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f9cc1dcd4fa71113fc6c4cfc35f94bf90c80f1dd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


x46085:
Hello and welcome. I'm curious was this an actual problem you ran into while using Elodin? Or just looking for an issue item to fix?

zxq82lm:
Thanks for the contribution, @Mohit-Ak! Great cleanup, and thanks for including the regression test. I’ll just add the corresponding documentation updates to your branch before we merge. Nice work! 👏

zxq82lm (approved):

PR #849, opened 2026-09-10
fixes for from scratch testing on mac

x46085:
Looking good! I was able to build and test with these tweaks.

I noticed that I wasn't able to maintain 8000hz, the logs were reporting ~5000/8000 per second. I was still able to fly it. The controls are not super smooth feeling, probably plenty of polish work could be done on the "feel" of the drone controls. But was impressive to see how stable it is.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the betaflight-sitl example build script and EEPROM bootstrap; no production flight or auth paths.
> 
> **Overview**
> Updates the **macOS two-pass SITL build** for Betaflight **2026.6.1**: stub objects live outside the SITL object tree, the link step uses the Makefile’s **`TARGET_OBJS`** list (instead of linking every `.o` under `obj/`), and stubs are narrowed/extended for unresolved symbols (`dmaGetHandlerCount`, `serialPinConfig_System`, lap-timer PG) while dropping redundant GPS shims.
> 
> **`init_eeprom.py`** gets process lifecycle handling (`stop_sitl`, stale `pgrep` cleanup, longer CLI wait, stderr on timeout) so from-scratch runs don’t leave zombies or hang on first-boot EEPROM programming. It also **disables `rc_smoothing`** during init/verify because high lockstep RX rates break auto cutoff and can zero the throttle channel.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1135c573e38237e471a9de59ba48fa7d090bcd7e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


anderspitman:
@x46085 what problem did you hit that required the changes to init_eeprom.py? I'm a little protective of keeping that file small because I already had to fight my agent to keep the original version simple.

Note that it's just a convenience script. You can still run the old commands manually if desired.

x46085:
> Note that it's just a convenience script. You can still run the old commands manually if desired.

Sure, but I prefer the README steps to run flawlessly as described. If it isn't expected to work, better to remove from the README. I didn't dig very hard on it, this is what the agent said:

"The ./build.sh link was failing on macOS because the Betaflight 2026.6.1 upgrade outgrew the old stub path. That is why this used to work on this Mac and then stopped.

Two things broke together:

New unresolved symbols. 2026.6.1’s DMA refactor requires dmaGetHandlerCount, and SITL still calls serialPinConfig() even though it does not compile serial_pinconfig.c.
The manual link was too greedy. find … -name '*.o' also picked up compile-only extras (dma_common.o, alternate autopilot PGs). After the missing symbols were filled in, that produced duplicate-symbol errors. The old GPS stubs also collided with the real io/gps.c that 2026.6.1 now compiles.
examples/betaflight-sitl/build.sh now stubs only the symbols the Makefile leaves unresolved and links the Makefile’s TARGET_OBJS plus those stubs.

The rebuild succeeded. I smoke-tested betaflight/obj/main/betaflight_SITL.elf; it started and bound 5761 for UART1."

x46085:
> > Note that it's just a convenience script. You can still run the old commands manually if desired.

Original error:

./build.sh             py elodin nix impure 12:07:36
SITL lockstep event patch already applied
Building Betaflight SITL...
  Target: SITL
  Gyro/PID rate: 8000 Hz (must match config.py simulation_rate)
  Run-loop delay: 0 us (0 = busy-wait, best lockstep latency)
  Output: /Users/danieldriscoll/dev/elodin/examples/betaflight-sitl/betaflight/obj/main/betaflight_SITL.elf

Note: ARM SDK not found. Installing (required for build system)...
make: Nothing to be done for 'arm_sdk_install'.
Building objects (first pass)...
Creating exe - copying obj/main/betaflight_SITL.elf to obj/betaflight_2026.6.1_SITL 
Compiling SITL stubs for macOS...
Linking with stubs...
Undefined symbols for architecture arm64:
  "_dmaGetHandlerCount", referenced from:
      _dmaAllocate in dma.o
      _dmaGetOwner in dma.o
      _dmaGetIdentifier in dma.o
  "_serialPinConfig_System", referenced from:
      _initPhase2 in init.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

x46085:
> @x46085 what problem did you hit that required the changes to init_eeprom.py?

Sorry those first to were related to the build.sh script regression, this was the impetus for the python script tweaks:

"The CLI timeout was a leftover SITL process. An earlier run had bound the UDP ports and entered the lockstep loop without opening UART1 on 5761, and init_eeprom.py kept waiting on that dead CLI instead of starting clean.

I updated the script so it now:

stops any existing betaflight_SITL.elf before starting
waits longer for first-boot EEPROM programming
kills the child and prints SITL stderr if the CLI never appears"

So this is mostly just hardening, catching a situation I ran into if you don't start things exactly the right way first time

PR #853, opened 2026-09-12
Add Package B FPV camera vertical slice for betaflight-sitl.

Zesterdock:
Opt-in RACE_CAMERA=1 registers drone.fpv, non-blocking latency reads, and editor sensor view.
This pull request completes "Package B" by adding an opt-in FPV camera to the Betaflight SITL example and updates the racing plan and code to reflect this milestone. The camera is off by default and can be enabled via the `RACE_CAMERA=1` environment variable. Camera frames are produced, consumed, and reported, with no impact on the default scripted run. The documentation and status tracking have been updated to mark Package B as complete and to provide verification evidence.

---

> [!NOTE]
> **Medium Risk**
> Adds optional GPU render-server startup and new hot-loop DB reads, but defaults preserve existing behavior; misconfiguration only fails fast on invalid RACE_CAMERA values.
> 
> **Overview**
> Completes **Package B** by adding an opt-in FPV camera to the betaflight-sitl example via **`RACE_CAMERA=1`**, while **`RACE_CAMERA=0` (default)** keeps the scripted SITL path unchanged and does not start the render server.
> 
> When enabled, the sim registers **`drone.fpv`** (640×360 RGBA @ 30 Hz, Section 7.4 contract), updates the editor schematic with **`sensor_view`** and a ground plane, and performs **non-blocking, latency-adjusted** `read_msg` at most once per camera period **after** the Betaflight lockstep motor exchange. Frames are passed into **`GuidanceUpdate`** (`frame`, `frame_sample_time`, `frame_fresh`) without changing scripted/manual guidance behavior. Shutdown prints first-frame timing, sample counts, and observed simulated FPS (with a ≥15 FPS acceptance check).
> 
> **`RACING_PLAN.md`** marks Package B complete, records 2026-09-11 verification (C0 still PASS, exportable MP4), and updates the resume point (D manual qualification; G unblocked).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f16a4fc61cc23d16385752fdbffee2bd64a0651a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


anderspitman:
@cursor review this

examples/betaflight-sitl/main.py
```diff
@@ -138,9 +159,53 @@ def cleanup_stale_betaflight():
...
 # Editor schematic for visualization
-world.schematic(
+if CAMERA_ENABLED:
+    _schematic = f"""
+    tabs {{
+        hsplit name = "Viewport" {{
+            viewport name=Viewport pos="drone.world_pos + (0,0,0,0, 10,10,5)" look_at="drone.world_pos" show_grid=#true show_frustums=#true active=#true
+            vsplit share=0.3 {{
+                sensor_view "{FPV_MSG}" name="FPV Camera"
+                graph "drone.motor_command" name="Motor Commands (from Betaflight)"
```
anderspitman, line 192: Would prefer not to duplicate the rest of the schematic here. We only want to inject the FPV camera if it's enabled.

`examples/falcon9/visual_check.py`, `examples/n-body/sim.py`, and `examples/sensor-camera/main.py` show some examples of ways to dynamically generate the KDL you need at runtime. You don't necessarily have to do exactly what those do, but they might help you get started.

anderspitman: @Zesterdock this comment hasn't been addressed

anderspitman (changes requested):
In addition to my inline comment below, here are some review comments from my agent:

# PR #853 Review Findings

## Verdict: Request changes

The camera vertical slice largely works: camera-enabled and disabled runs pass, valid 640×360 RGBA frames arrive at approximately 30 simulated FPS, C0 remains healthy, video export succeeds, and failed camera acceptance now returns a nonzero exit status.

The following issues remain:

### 1. Held frames are marked as fresh

In `examples/betaflight-sitl/main.py`, `_read_fpv_frame` sets `latest_fresh = True` for every valid nominal read, even when `read_msg_at` returns the same selected timestamp as the previous read.

A repeated selected timestamp is sample-and-hold, not a newly produced renderer frame. At the accepted minimum of 15 renderer FPS, downstream perception could therefore process each frame twice. `frame_fresh` should only be true when `selected_ts` changes; a held frame may still be supplied with `frame_fresh=False`.

### 2. Shutdown frame accounting is not exact

The shutdown report does not provide the total number of distinct renderer frames. In the review run it reported:

- `sample_count=449` (successful nominal reads, potentially including held frames)
- `unique_selected_timestamps=387` (post-warmup measurement window only)
- exported video: **451 actual frames**

The timestamp sweep can also include a floor-selected frame from before the warmup boundary and miss a final frame after its last cursor. Selected timestamps should be explicitly bounded to the measurement interval, and the report should distinguish total renderer frames from post-warmup frames used for FPS calculation.

## Architecture recommendation

The highest-value maintainability improvement would be to move the FPV runtime logic out of the already-large `main.py` into a focused module such as `fpv_camera.py`. That module could own sampling, freshness, statistics, and acceptance, leaving `main.py` responsible only for orchestration and constructing `GuidanceUpdate`. This would provide a clean, independently testable seam for the later perception packages without requiring a broader redesign.

## Test gap

No focused tests cover duplicate selected timestamps, accounting-window boundaries, malformed frames, or camera acceptance. These timestamp and acceptance rules can be tested without a GPU and should have pure regression coverage.

x46085:
btw it's super helpful to see a little video in the PR of the feature in action when it's visual, here's PR from Ben showing best practice: https://github.com/elodin-sys/elodin/pull/865

PR #861, opened 2026-09-18
fix: Editor tick step.

### LINKED ISSUE #834: Paused Editor Playback, advance by tick with arrows no-op
We used to be able to advance by one tick at a time, using the arrows and the GUI. Seems to be broken now.

zxq82lm:
> [!NOTE]
> **Medium Risk**
> Changes playback stepping, async query lifecycle on reconnect, and how simulation rate is derived—core timeline behavior with broad test coverage but non-trivial edge cases around stale handlers.
> 
> **Overview**
> Fixes timeline tick stepping and TPS when a recording does not publish `simulation_time_step`, and stops stale rates from surviving reconnects or recording switches.
> 
> **Simulation rate resolution** (`impeller2_bevy`): Fetches declared `Globals.simulation_time_step` with paced retries and generation guards; seeds `SimulationTimeStep` at **0** (unknown) instead of a fake 1 ms default (#834). When nothing is declared, the editor estimates the finest mean sample spacing from component min/max/count queries. `rearm_sim_time_step` clears resolution state and the published step on `NewConnection` or `DbConfig` identity change.
> 
> **Time-range queries** (`data_overview`): Phase‑1 SQL batches move to `dispatch_time_range_queries` so they run without the Data Overview panel open; `ComponentTimeRanges::reset()` plus generation stamps ignore in-flight replies after a reset.
> 
> **Timeline UX**: Frame step/back prefer the next/previous **cached sample** on subscribed (high-priority) series when coverage allows, with nominal `SimulationTimeStep` as fallback. Status bar shows `~TPS` and tooltips for estimated vs declared rates. Timeline slider no longer takes an FPS parameter for labels.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d10ce688f6c12cc518ab035095e4c5b94b6f907. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


x46085:
Make sure to have verified with high-rate databases (1000hz) both a real world and simulation (falcon9 is 1000hz)

libs/impeller2/bevy/src/lib.rs
```diff
@@ -302,6 +338,39 @@ pub struct SeriesFetchPriority {
...
 
+/// Nearest recorded sample boundary after `after` across the subscribed
+/// components — the next moment something the user is actually looking at
+/// changes, which is what a frame step should land on.
+///
+/// A component whose coverage cannot vouch for the gap is skipped rather than
+/// vetoing the step: the answer is still a real boundary of a displayed series,
+/// which beats a nominal step. `None` when nothing can answer, leaving the
+/// caller to fall back.
+pub fn next_subscribed_sample(
```
anderspitman, line 349: Comment from my agent:


### Frame stepping may skip a nearer known sample when cache coverage differs

In `libs/impeller2/bevy/src/lib.rs:341-371`, `next_subscribed_sample` and `prev_subscribed_sample` omit a subscribed component whenever coverage cannot prove that its cached neighbor is the immediately adjacent recorded sample. This can cause the selected boundary from another component to be farther away than a sample already known to exist.

For example, while progressive backfill is incomplete, a high-rate gyro series might have a known sample at `9.999 s` without coverage for the whole gap, while a low-rate GPS series is covered and has its previous sample at `9.000 s`. Stepping backward from `10.000 s` discards the gyro candidate and jumps to the GPS sample, skipping the known gyro change. Because a resolved sample boundary takes precedence over the nominal time step, this can also produce a larger jump than the previous fixed-step behavior.

This is most relevant to long, multi-rate recordings while their subscribed series have progressed to different coverage points. A mixed covered/uncovered-series test would clarify the intended behavior. Should an uncovered component's known nearer boundary cap the result, or should mixed coverage cause the code to fall back to the nominal step until the gap is fetched?

anderspitman (approved):
Approved, pending agent comment below and assuming 1kHz verification is completed.

libs/impeller2/bevy/src/lib.rs
anderspitman: Agent managed to capture a video repro:


https://github.com/user-attachments/assets/4dbd4747-84d3-481f-add4-beca4d0850d0

anderspitman: Note I'm not sure how much of an issue this would be in practice, which is why I didn't request changes. It might be fine.

zxq82lm: That gives me a good opportunity to take a closer look at it.
Thanks for the review!

zxq82lm:
> Make sure to have verified with high-rate databases (1000hz) both a real world and simulation (falcon9 is 1000hz)

- db with metadata (sim time step):
image

- db without metadata (telemetry-like):
image

